PR #81, opened 2019-12-06
remove hard-coded RabbitMQ password

LincolnBryant:
I've replaced the hard-coded `leftfoot1` value with the one specified in the Values file for RabbitMQ.

### DIFF
--- a/servicex/templates/did-finder-deployment.yaml
+++ b/servicex/templates/did-finder-deployment.yaml
@@ -24,7 +24,7 @@ spec:
       - name: {{ .Release.Name }}-did-finder
         image: {{ .Values.didFinder.image }}:{{ .Values.didFinder.tag }}
         command: ["bash","-c"]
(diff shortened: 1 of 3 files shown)


gordonwatts:
That password actually came from me long ago. The reason for it was that the password came in via a secret and that was painful when I wanted to look myself at the web API. So I froze it as then it was easy to just do a port-forward. Does this keep that feature, or does this mean we have to go back to several steps to bring up the web docs?

cnweaver:
The only thing that this patch changes is to make the settings in the chart's values file actually work as advertised. I don't think that this would have any impact on port forwarding, and in general it doesn't stop you from using the password, you just have to know what you specified when configuring the chart rather than assuming the default.

LincolnBryant:
By default, the docs and chart assume the password is `leftfoot1`. As Chris notes, this simply makes the chart respect the configurable in the values file if the user chooses to change it.

BenGalewsky:
This PR is great, but I let it go stale. superseded by PR #106

PR #115, opened 2020-02-20
Add config for RabbitMQ retries

BenGalewsky:
# Problem
The App has been updated to expect configuration settings for connecting to RabbitMQ Server.

# Approach
- Added these settings to the app.conf template
- Added new values to the chart. Defaults to two minutes of retries
- Updated README


README.md
```diff
@@ -180,6 +180,8 @@ or [mino](https://github.com/helm/charts/tree/master/stable/minio#configuration)
 | `app.image`                          | ServiceX_App image name                          | `sslhep/servicex_app`                                   |
 | `app.tag`                            | ServiceX image tag                               | `latest`                                                |
 | `app.pullPolicy`                     | ServiceX image pull policy                       | `IfNotPresent`                                          |
+| `app.rabbitmq.retries`               | Number of times to retry connecting to RabbitMQ on startup | 12                                            |
```
mweinberg2718: From the app.conf.template it looks like this should actually be 24 tries, each after 5 seconds

Issue #80, opened 2019-12-06
Remove GRID cert sensitive information from yaml files, and use a k8 secret.


gordonwatts:
# Story
As a ServiceX Administrator I want to keep the GRID security certificates as a separate k8 secret so I don't need to cache the private grid key cert and password in a yaml text file sitting around insecurely on my computer.

# Assumptions
1. Trying to access a file from ATLAS (different story from CMS in case their access and cert system is difference?)

# Acceptance Criteria
1. Admin of ServiceX can create a k8 secret using instructions from ServiceX docs
1. No GRID information is required to be in a yaml file used to start the service
1. ServiceX can access a GRID dataset from ATLAS (e.g. the Zee dataset in the demo notebook).

gordonwatts:
See [`func_adl_server`](https://github.com/gordonwatts/func_adl_server) for one possible approach. Search for the word "secret" in the page to see the instructions I used for building the secret. And the [`rucio` pod](https://github.com/gordonwatts/func_adl_server/blob/master/func-adl-server/templates/as-rucio-downloader.yaml) that mounts the secret info here for how I injected the information into the service.

BenGalewsky:
Could we include the certs still in values.yaml, but the user has to create a shell inside the x509_secret pod to enter their password the first time? If there is a script run on the user's desktop then we have to figure out how to distribute that and make sure they have the dependencies

gordonwatts:
@BenGalewsky of course that is possible. In general this means standing this thing up repeatedly is quite painful. My approach was to not create a yaml file for the secret, but instead use the command line to create the secret on the fly. The command line pulled from the `userkey` and `usercert` files, and then had values for user name, VOMS, and password. So the password did get stored as a secret...

My goal with the above design was that after installing the secret with a separate command line, I could just run the helm chart, update it, uninstall it, and reinstall it and it would continue to work without me constantly having to enter credentials.

Something for us not to do: sweat over this too long. The whole infrastructure is being replaced. Anytime you hear anyone from DOMA talking about "replacing globus", they are talking about changing how security and identity are done on the net. So whatever we have now is what we need for the next year or perhaps two, but after that everything is going to change. I believe towards tokens, so we probably know the shape of the new system.

PR #116, opened 2020-02-24
Update x509 deployment to mount secrets generated by servicex cli

BenGalewsky:
# Problem
The current way of providing grid certs and passphrase to the x509 proxy requires putting them into the helm chart's `values.yaml` - this works, but is not very secure.
Partial solution to #80 

Also, when you delete the helm chart the x509 proxy hangs around.

# Approach
The new servicex-cli has a command for putting the certs and passphrase into a secret. In this
PR we change the x509 deployment to mount this secret instead of the old grid certs ConfigMap and passphrase secret.

Also, to insure the x509 proxy secret is deleted when the helm chart is deleted, we add the creation of an empty x509-secret in the helm chart so it will be cleared upon delete. The X509 secret service will add files to it.

Finally, updated the README to reflect this new configuration process


BenGalewsky:
@cnweaver - the secret with the cert and passphrase are now created with the cli script [here](https://github.com/ssl-hep/servicex_cli/blob/master/src/servicex/servicex.py)

cnweaver:
I assume you mean https://github.com/ssl-hep/servicex_cli/blob/develop/src/servicex/servicex.py , and, yes, I looked at that for reference. It sounds like that means templates/x509-proxy-secret.yaml is indeed unnecessary?

BenGalewsky:
Ah, I get it - `x509-proxy-secret.yaml` is in the helm chart so that a skeletal secret is deployed with the chart. That means when you delete the chart, the proxy that is created by the x509-secret service also gets deleted.
I added a comment to the yaml file to clarify

cnweaver:
Okay, that's a clever trick. However, it still looks like the name of the secret doesn't match, `{{ .Release.Name }}-x509-proxy` vs. `grid-certs-secret`.

Issue #146, opened 2020-06-29
ServiceX Transform Response Should Provide Minio URL


BenGalewsky:
_As an analyzer I want to know the Minio URL so I can download my results_

## Assumptions
1. Added to app.conf
2. Provided as part of the status response

## Acceptance Criteria
1. Deploy a serviceX with a public Minio Endpoint
2. Submit a request
3. See the URL for Minio in the status response
4. Able to access via browser

gordonwatts:
I just wanted to say: YES. I've been wanting to see this for a while. This is something that will have to be fed from the helm chart. But this is great.

Please do make sure to let me know as soon as it is understood how this will work so that we can start to understand how to put this into the frontend code.

PR #147, opened 2020-07-06
Add public URL to app config

BenGalewsky:
# Problem
A client of serviceX needs to know the URL, username, and password of the Minio instance where results are stored.

Partial solution to ServiceX issue #146

# Approach
Add `MINIO_PUBLIC_URL` to the flask app ConfigMap. If no ingress is requested, then provide the internal service name so it works in Jupyter.

### DIFF
--- a/servicex/templates/flask_app-configmap.yaml
+++ b/servicex/templates/flask_app-configmap.yaml
@@ -67,6 +67,11 @@ data:
     MINIO_URL_TRANSFORMER = '{{ .Release.Name }}-minio:9000'
     MINIO_ACCESS_KEY = '{{ .Values.minio.accessKey }}'
     MINIO_SECRET_KEY = '{{ .Values.minio.secretKey }}'
+    {{ if .Values.minio.ingress.enabled }}
+    MINIO_PUBLIC_URL = '{{index .Values.minio.ingress.hosts 0 }}'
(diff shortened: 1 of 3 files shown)


servicex/templates/flask_app-configmap.yaml
```diff
@@ -67,6 +67,11 @@ data:
     MINIO_URL_TRANSFORMER = '{{ .Release.Name }}-minio:9000'
     MINIO_ACCESS_KEY = '{{ .Values.minio.accessKey }}'
     MINIO_SECRET_KEY = '{{ .Values.minio.secretKey }}'
+    {{ if .Values.minio.ingress.enabled }}
+    MINIO_PUBLIC_URL = '{{index .Values.minio.ingress.hosts 0 }}'
+    {{ else }}
+    MINIO_PUBLIC_URL = '{{ .Release.Name }}-minio:9000'
```
gordonwatts, line 73: Should this internal name be the name of the service? And shouldn't that be stored in a macro somewhere, rather than generated here? Is the port 9000 settable?

BenGalewsky: I made it possible to override if you don't want this handy default. The 9000 is set inside the Minio sub chart, so now picking it up from that chart's values

PR #158, opened 2020-07-14
Add Slack webhook URL for new signups to app configmap

### LINKED ISSUE #152: Add Slack webhook for signup alerts
# Story
As a ServiceX admin I want to be notified when a new user signs up for the service.

The current authentication method is sufficient but bare-bones; in addition it would be good to have a Slack webhook or mailgun to alert admins when a new person signs up

AndrewEckart:
This adds an optional secret to the Flask app, which is then used to populate environment variables (namely `SLACK_WEBHOOK_URL`). The name of the secret can be specified in `values.yaml` (default null so that the app can be deployed without one).

Related to ssl-hep/ServiceX_App#35, which partially resolves #152.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,8 +190,8 @@ or [mino](https://github.com/helm/charts/tree/master/stable/minio#configuration)
 | `app.ingress.enabled`                | Enable install of ingres                         | false                                                   |
 | `app.ingress.host`                   | Hostname to associate ingress with               | uc.ssl-hep.org                                          |
 | `app.ingress.defaultBackend`         | Name of a service to send requests to internal endpoints to | default-http-backend                         |
-| `app.resources`                      | Pass in Kubernetes pod resource spec to deployment to change CPU and memory | { }                          |
-| `app.secret`                         | Name of a Kubernetes Secret which is used to populate environment variables |                              |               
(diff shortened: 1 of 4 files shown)


BenGalewsky:
Can you update the README to show new value that was added and describe how this is configured?

README.md
```diff
@@ -190,7 +190,8 @@ or [mino](https://github.com/helm/charts/tree/master/stable/minio#configuration)
 | `app.ingress.enabled`                | Enable install of ingres                         | false                                                   |
 | `app.ingress.host`                   | Hostname to associate ingress with               | uc.ssl-hep.org                                          |
 | `app.ingress.defaultBackend`         | Name of a service to send requests to internal endpoints to | default-http-backend                         |
-| `app.resources`                      | Pass in Kubernetes pod resource spec to deployment to change CPU and memory | { } |               
+| `app.resources`                      | Pass in Kubernetes pod resource spec to deployment to change CPU and memory | { }                          |
+| `app.secret`                         | Name of a Kubernetes Secret which is used to populate environment variables |                              |
```
BenGalewsky, line 194: Since the secrets->Environment Var mapping is hardcoded in the app_deployment, you should start a new section in the README that documents which secrets are supported and what they do

PR #167, opened 2020-07-27
Make autoscaler CPU limit configurable

mweinberg2718:
Updates README, Flask configmap, and values.yaml to accommodate new CPU threshold configuration for autoscaling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ or [mino](https://github.com/helm/charts/tree/master/stable/minio#configuration)
 | `transformer.pullPolicy`             | Pull policy for transformer pods (Image name specified in REST Request) | IfNotPresent |
 | `transformer.autoscalerEnabled`      | Set to True to enable the pod horizontal autoscaler for transformers |  False          |
 | `transformer.cpuLimit`               | Set CPU resource limit for pod in number of cores | 1 |
-| `transformer.cpuScaleThreshold`      | Set CPU percentage threshold for pod scaling | 1 |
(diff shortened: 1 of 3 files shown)


servicex/values.yaml
```diff
@@ -78,6 +78,8 @@ codeGen:
 transformer:
   pullPolicy: IfNotPresent
   autoscalerEnabled: false
+  cpuLimit: 1
+  cpuScaleThreshold: 1
```
BenGalewsky, line 82: Is this a good default value for people to use?

mweinberg2718: Maybe not for the cpuScaleThreshold. After some testing on high-efficiency transformers, I would suggest that 70% provides a reasonable balance between speed of spinning up a new pod and the issue of making too many unused pods.

README.md
```diff
@@ -224,6 +224,8 @@ or [mino](https://github.com/helm/charts/tree/master/stable/minio#configuration)
 | `minio.ingress.hosts`                | List of hosts to associate with ingress controller | nil |
 | `transformer.pullPolicy`             | Pull policy for transformer pods (Image name specified in REST Request) | IfNotPresent |
 | `transformer.autoscalerEnabled`      | Set to True to enable the pod horizontal autoscaler for transformers |  False          |
+| `transformer.cpuLimit`               | Set CPU resource limit for pod in number of cores | 1 |
+| `transformer.cpuScaleThreshold`      | Set CPU percentage threshold for pod scaling | 1 |
```
BenGalewsky, line 228: Please fix this to match `values.yaml`

mweinberg2718: Good catch!

PR #187, opened 2020-08-19
Monitor services with Prometheus

AndrewEckart:
Our goal is to monitor each ServiceX microservice using Prometheus. This PR enables monitoring of RabbitMQ, Minio, and PostgreSQL using ServiceMonitors which are enabled via their respective helm charts.

It assumes that the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator) is installed on the cluster (possibly with [its helm chart](https://github.com/helm/charts/tree/master/stable/prometheus-operator)), and that it is configured to discover ServiceMonitors in whichever namespace ServiceX is deployed in. If both Prometheus and ServiceX are in the default namespace, this can be achieved by setting `prometheus.prometheusSpec.serviceMonitorSelectorNilUsesHelmValues` to false in a custom values.yaml file when deploying Prometheus.

Monitoring for our own services, such as the DID finder, transformers, and API server, can be added in separate PRs.


ivukotic:
Since we gave up on Prometheus, we can close this.

PR #189, opened 2020-08-21
Organize templates into subfolders

AndrewEckart:
Since we already have a fair number of templates, and that number is likely to grow in the near future with the addition of a few ServiceMonitors, this PR reorganizes the existing templates into subdirectories by microservice.

This is somewhat uncommon but makes sense for charts with a large number of templates. Here are two examples:
- [Prometheus](https://github.com/helm/charts/tree/master/stable/prometheus/templates) - organized by resource type (deployments, configmaps, ...)
- [Prometheus Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator/templates) - organized by microservice (prometheus, grafana, ...).

I went with the latter approach, but would be happy with either one.


BenGalewsky:
I like this - good choice organizing by Microservice, that makes sense to me

PR #197, opened 2020-09-23
Add docs URL to app configmap

AndrewEckart:
We want to make sure that each deployment points to the correct version of the documentation. For example, an RC3 deployment might point to https://servicex.readthedocs.io/en/v1.0.0-RC3/ (does not exist atm), while a deployment made from the develop branch (perhaps via CD) should point to https://servicex.readthedocs.io/en/latest/.

The easiest way to do this is to simply add another config value to provide the API server with the base URL of the docs.


AndrewEckart:
Refactored to use chart version in `servicex/Chart.yaml`.

This creates a tight coupling between release branch names (which are also RTD versions) and chart versions. On any given release branch `v1.x.x`, the version property in Chart.yaml *must* be set to `version: 1.x.x`.

PR #211, opened 2020-10-19
Update user documentation

AndrewEckart:
This PR contains the new getting started guide, "ServiceX in 5 minutes", which aims to bring new users from zero to making their first request and seeing some data in the terminal in the minimum number of possible steps. It adds sections to the RtD navbar.

It replaces the old getting started guide, which only had information on how to obtain credentials (no example requests), and had not been updated since RC1.

While the new guide reflects the **current state of RC3**, there are a few unresolved issues:
- Can we have a button to download a `.servicex` file and streamline this process even further? See ssl-hep/ServiceX_App#70
- Do users need to install both `servicex` and `func-adl-servicex`? Is @gordonwatts going to create an umbrella package which contains the `servicex` client and all relevant query languages? Whenever this is resolved, the installation instructions in this guide should be updated.
- We probably want examples for every common (file type, query language) combination. Currently there are only examples for (xAOD, func_adl), and (uproot, func_adl). We could use an example of (uproot, TCut), at minimum. See also: ssl-hep/ServiceX_frontend#122.
- I was only able to get the uproot example to work by specifying `image="sslhep/servicex_func_adl_uproot_transformer:issue6"`. We should remove this from the documentation as soon as the default images for `v1.0.0-rc.3` and `develop` are working.
- Even in the examples I provided, there is a heavy ATLAS bias. Can @mweinberg2718 provide some examples for CMS?


AndrewEckart:
Updated to reflect that as written, the guide only works with:
- Python 3.6 or 3.7 (3.8 support is blocked by func-adl)
- `servicex==2.1b4`
- `func-adl-servicex==1.0a4` - in 1.0a5, ServiceXDatasetSource is replaced with ServiceXSourceUpROOT and ServiceXSourceXAOD. I can update the version and the example code as soon as the frontend settles on a stable version.

gordonwatts:
Should be `func-adl-servicex==1.0` - released a new version yesterday: https://pypi.org/project/func-adl-servicex/

BenGalewsky:
When I try to run the xaod example in a clean virtualenv after `pip install func-adl-servicex==1.0`

I get:
```
    from func_adl_servicex import ServiceXDatasetSource
ImportError: cannot import name 'ServiceXDatasetSource' from 'func_adl_servicex' (/Users/bengal1/.virtualenvs/func_adl/lib/python3.7/site-packages/func_adl_servicex/__init__.py)
```

AndrewEckart:
> When I try to run the xaod example in a clean virtualenv after `pip install func-adl-servicex==1.0`
> 
> I get:
> 
> ```
>     from func_adl_servicex import ServiceXDatasetSource
> ImportError: cannot import name 'ServiceXDatasetSource' from 'func_adl_servicex' (/Users/bengal1/.virtualenvs/func_adl/lib/python3.7/site-packages/func_adl_servicex/__init__.py)
> ```

See above:
> in 1.0a5, ServiceXDatasetSource is replaced with ServiceXSourceUpROOT and ServiceXSourceXAOD

There are current examples at
https://github.com/iris-hep/func_adl_servicex/tree/f108f3a5f1ad9aced97c511cb3881025cb62e0fe/scripts

I'll update this PR later today.

AndrewEckart:
This has been updated with `func-adl-servicex==1.0` installation instructions and syntax. I've also changed the uproot endpoint from uproot.servicex.ssl-hep.org to uproot-atlas.servicex.ssl-hep.org.

AndrewEckart:
Okay, this has been updated once more with the latest frontend changes. We now ask users to `pip install servicex-clients`, which is the umbrella package containing all of the necessary libraries.

I've also cleaned up the requests guide in `requests.md`, which covers most of the basics of using func-ADL. I tested all of the example requests against the current xAOD backend on Ubuntu 20.04 with Python 3.8, with good results. Finally, I added instructions on how to bypass the local cache.

The docs are not comprehensive - we are still missing information about how to make requests to uproot backends, how to specify requests with TCut, etc. and we would also like to embed Gordon's Jupyter notebooks within the docs. However, they are greatly improved, so this is a good foundation to build on.

PR #219, opened 2020-10-31
Generate random JWT secret

BenGalewsky:
#Problem
The JWT Secret and flask App Secret are hardcoded in the app config template.

# Approach
Changed these both to randomly generated values


BenGalewsky:
Make the secret configurable in values.yaml

PR #221, opened 2020-11-06
Add DISABLE_USER_MGMT config value to Flask app

AndrewEckart:
See ssl-hep/ServiceX_App#73.


AndrewEckart:
Re: the chart, this would be a good opportunity to reorganize the values a little bit, into something more like:

```
app:
  auth:
    enabled: true
    disableUserMgmt: false (default)
    globusClientID: *****
    globusClientSecret: *****
```
I think that the `mailgunApiKey`, `mailgunDomain`, and `slackSigningSecret` should be kept at the `app` level since those could conceivably be used for other things. I'm not sure about the `newSignupWebhook`, though, since that only applies when auth and user management are enabled.

Issue #230, opened 2020-12-07
ingress settings


ivukotic:
Application ingress points to three services:
  xaod.servicex.ssl-hep.org
                             /minio               xaod-minio:9000 (192.168.38.190:9000)
                             /servicex/internal   default-http-backend:80 (<error: endpoints "default-http-backend" not found>)
                             /                    xaod-servicex-app:8000 (192.168.48.136:5000)

it should not point to minio as minio has its own ingress.
it should not point to a default backend as there is no service listening on port 80.

BenGalewsky:
Removed the unused Ingress path via PR #236 

The reference to the `default-http-backend` is a trick to hide the `/internal` path from the outside. Anyone attempting to access the internal endpoints via this ingress will get an nginx 503 error.

BenGalewsky:
Also, in case anyone is wondering why we use the Minio sub chart's ingress instead of adding it as a path here, the [Minio SDK](https://docs.min.io/docs/python-client-api-reference.html) only accepts a hostname and port to connect to the service. As per the S3 standard, there is no option to use an http path.

AndrewEckart:
Resolved in #236.

PR #236, opened 2020-12-10
Remove unused servicex ingress to minio

BenGalewsky:
# Problem
As pointed out by @ivukotic in ServiceX Issue #230 there is an unused ingress path to Minio from the serviceX ingress. This is not needed since we enable access to this service via the Minio helm chart.

# Approach 
Deleted the ingress path

### DIFF
--- a/servicex/templates/app/ingress.yaml
+++ b/servicex/templates/app/ingress.yaml
@@ -22,12 +22,6 @@ spec:
   - host: {{ .Release.Name }}.{{ .Values.app.ingress.host }}
     http:
       paths:
-      {{- if .Values.objectStore.enabled }}
-      - path: /minio
-        backend:
-          serviceName: {{ .Release.Name }}-minio
-          servicePort: 9000
-      {{- end }}
       - path: /servicex/internal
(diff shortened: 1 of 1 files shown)


ivukotic:
@Benjamin Galewsky<mailto:bengal1@illinois.edu>, if there is no good reason for a dedicated minio ingress,  it would be simpler to just reuse existing app ingress and have minio serve on a subdomain.

My 2c,
Ilija

BenGalewsky:
@ivukotic - see comments on Issue #230 as to why we don't use a subpath on the serviceX ingress

ivukotic:
@Benjamin Galewsky<mailto:bengal1@illinois.edu> You can make it serve on a path of your choosing by adding path_prefix  in its configuration?
Then just prefix objects?

Ilija

AndrewEckart (approved):
It'd be nice to have just one Ingress, which would make deployment simpler, but it sounds like that's blocked by upstream limitations.

Removing it from the develop branch eliminates confusion - we can experiment on feature branches in the future if needed.

Issue #238, opened 2020-12-11
Autoscaling


ivukotic:
* Autoscaling should be turned on by default.
* Users should not be able to request number of CPUs
* targeted CPU utilization should be set to 0.2 or 0.3 since transformers are quite inefficient.
* documentation should be updated to clarify that the k8s cluster need to be able to run eg. kubectl top nodes

robrwg:
Wondering if there is a blocker here of some sort, or perhaps was this already accomplished?

PR #241, opened 2020-12-22
Enable autoscaling by default, set threshold to 30%

AndrewEckart:
Addresses #238.

The described changes are implemented, but won't this create a hard dependency on the horizontal pod autoscaler (HPA)? If so, we should document that it must be installed on your cluster as a prerequisite for deploying ServiceX.


gordonwatts:
This should get tested with both xaod and uproot backends which have very different performance characteristics. It would also be good to use @d-w-liu 's stress test on this as it will try to process 100's of files.

ivukotic:
I tested it with both xaod and uproot transformers (just examples from the documentation) and both work as expected. Given it's already month and a half this is waiting, I will merge it.

PR #244, opened 2021-01-13
update documentation to use servicex.yaml fix heml repo

ivukotic:
docs and repository fix


AndrewEckart:
I think these should be two separate PRs. The doc changes can be merged right away but the chart requirements fix needs to be tested. I don't think we should use Minio chart version 5.0.*, because we have had issues in the past with a bug that was not fixed until chart version 5.0.16 or something like that.

ivukotic:
Reverted back repo changes. Will add them back in a different PR.

AndrewEckart (approved):
Thanks for updating the docs to reflect the changes in ssl-hep/ServiceX_App#72 and ssl-hep/ServiceX_frontend#138!

Issue #215, opened 2020-10-29
"can't give argument to non-function" error while installing servicex locally


sudo-panda:
I am using minikube to run a Kubernetes cluster on my PC.

When using the command :

```
helm install -g -f demo-values.yaml ssl-hep/servicex
```

I get the following error:

```
Error: template: servicex/charts/minio/templates/deployment.yaml:198:20: executing "servicex/charts/minio/templates/deployment.yaml" at <(not .Values.gcsgateway.enabled) (not .Values.azuregateway.enabled) (not .Values.s3gateway.enabled) (not .Values.b2gateway.enabled)>: can't give argument to non-function not .Values.gcsgateway.enabled
```

Is there any known solutions to this error.

AndrewEckart:
Thanks for the report! Here's my summary of this issue.

#### Cause
This is an upstream error in the Minio chart:
https://github.com/helm/charts/issues/21446

It was fixed in Minio 5.0.16 (March 17, 2020):
https://github.com/helm/charts/pull/21447

However, this fix won't be applied if your local version of the chart is out of date.

#### Temporary fix for @sudo-panda 

To fix it, try running `helm repo update`. 

To confirm that the update worked, run `helm show chart stable/minio` and check that the version is 5.0.16 or later:
```bash
$ helm show chart stable/minio
apiVersion: v1
appVersion: master
deprecated: true
description: DEPRECATED MinIO is a high performance data infrastructure for machine
  learning, analytics and application data workloads.
home: https://min.io
icon: https://min.io/resources/img/logo/MINIO_wordmark.png
keywords:
- storage
- object-storage
- S3
name: minio
sources:
- https://github.com/minio/minio
version: 5.0.33
```
You should now be able to make deployments without running into this issue.

#### Permanent fix (@BenGalewsky)
To permanently fix this, we need to (at minimum) update the ServiceX chart dependencies in `requirements.yaml` to require version 5.0.16+ of the Minio chart.

On a related note, the Minio developers have migrated their chart from the [stable](https://github.com/helm/charts) Helm repo (which has been deprecated by Google) to their own Helm [repo](https://github.com/minio/charts) (called `minio`) at https://helm.min.io/.

[...]

PR #245, opened 2021-01-14
update to helm dependences

ivukotic:
just moving to a working repository

### DIFF
--- a/servicex/requirements.lock
+++ b/servicex/requirements.lock
@@ -1,12 +1,12 @@
 dependencies:
 - name: rabbitmq
   repository: https://charts.bitnami.com/bitnami/
-  version: 7.6.8
+  version: 7.6.*
 - name: minio
   repository: https://helm.min.io/
   version: '>=5.0.32'
 - name: postgresql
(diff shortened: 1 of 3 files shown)


AndrewEckart (approved):
This PR addresses the [deprecation](https://helm.sh/blog/charts-repo-deprecation/) of the https://github.com/helm/charts repo (aka "stable") by Google. PostgreSQL has migrated to https://github.com/bitnami/charts/ (as RabbitMQ did), so our requirement needs to point there instead. 

Minio has migrated to its own chart repo, https://github.com/minio/charts, which has only 5.0.32 (they never published 5.0.33 to this repo but they have published new several major chart versions and are up to 8.x.x, which is a separate issue - see #247). Requiring `>=5.0.32` should avoid any breaking changes while also preventing the chart bug fixed in 5.0.16 (see #215).

LGTM. Thanks Ilija!

BenGalewsky:
This PR contains an unrelated change for the DID finder container. I moved the commits to a local branch and removed the incorrect commit. It is now PR #255 and #256

Issue #242, opened 2020-12-22
Chart should be versioned separately from microservices


AndrewEckart:
# Story
_As a ServiceX developer  I charts to be deployed regularly and correctly versioned so I can pick up new changes_

# Assumptions
1. Chart and application versions should be dissociated
2. Every change to the chart should result in a new patch version

# Acceptance Criteria
1. Users can pick up new charts with `helm repo update`

PR #246, opened 2021-01-20
Split app version from chart version

BenGalewsky:
# Problem
It doesn't make sense for the chart version to be tied to the microservice versions. It makes it hard to push new updates to the repo.

# Approach
Make `Chart.Version` be it's own semantic version stream, and tie `Chart.AppVersion` to the micro services

Solution to #242

### DIFF
--- a/servicex/templates/app/configmap.yaml
+++ b/servicex/templates/app/configmap.yaml
@@ -17,7 +17,7 @@ data:
 
     # Base url of documentation - autopopulated from chart version
     # Must be activated at https://readthedocs.org/projects/servicex/versions/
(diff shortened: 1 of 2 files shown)


servicex/templates/app/configmap.yaml
```diff
@@ -17,7 +17,7 @@ data:
 
     # Base url of documentation - autopopulated from chart version
```
AndrewEckart, line 18:
```suggestion
    # Base url of documentation - autopopulated from app version
```

AndrewEckart (approved):
LGTM!

PR #258, opened 2021-01-27
Changed all instances of IfNotPresent to Always.

ivukotic:
Since this is not expensive (basically just a hash check of images) we should have it as default.

### DIFF
--- a/river-xaod-values.yaml
+++ b/river-xaod-values.yaml
@@ -25,8 +25,8 @@ didFinder:
   site: mwt2
 
 objectStore:
-    enabled: true
-    publicURL: xaod-minio.servicex.ssl-hep.org
+  enabled: true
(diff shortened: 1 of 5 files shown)


river-xaod-values.yaml
```diff
@@ -80,6 +80,5 @@ rabbitmq:
       storageClass: rook-ceph-block
 transformer:
   autoscalerEnabled: false
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
```
BenGalewsky, line 83: This is very expensive, no? If I launch 200 workers then I do 200 pulls

ivukotic:
Hi,

Not really. It just compares has of the local image with the hash in dockerhub and does nothing if these are the same.

Ilija

PR #259, opened 2021-01-28
needs X509_CERT_DIR set.

ivukotic:
the new version really check cert.

### DIFF
--- a/servicex/templates/did-finder/deployment.yaml
+++ b/servicex/templates/did-finder/deployment.yaml
@@ -17,7 +17,7 @@ spec:
       - name: {{ .Release.Name }}-did-finder
         image: {{ .Values.didFinder.image }}:{{ .Values.didFinder.tag }}
         command: ["bash","-c"]
-        args: ["/usr/src/app/proxy-exporter.sh & sleep 5 && env PYTHONPATH=./src python3 scripts/did_finder.py --rabbit-uri amqp://user:{{ .Values.rabbitmq.auth.password }}@{{ .Release.Name }}-rabbitmq:5672/%2F {{ if .Values.didFinder.site }} --site {{ .Values.didFinder.site }} {{ end }} --threads {{ .Values.didFinder.threads }}"]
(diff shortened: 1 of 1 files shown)


servicex/templates/did-finder/deployment.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       - name: {{ .Release.Name }}-did-finder
         image: {{ .Values.didFinder.image }}:{{ .Values.didFinder.tag }}
         command: ["bash","-c"]
-        args: ["/usr/src/app/proxy-exporter.sh & sleep 5 && env PYTHONPATH=./src python3 scripts/did_finder.py --rabbit-uri amqp://user:{{ .Values.rabbitmq.auth.password }}@{{ .Release.Name }}-rabbitmq:5672/%2F {{ if .Values.didFinder.site }} --site {{ .Values.didFinder.site }} {{ end }} --threads {{ .Values.didFinder.threads }}"]
+        args: ["/usr/src/app/proxy-exporter.sh & sleep 5 & export X509_CERT_DIR=/etc/grid-security/certificates && env PYTHONPATH=./src python3 scripts/did_finder.py --rabbit-uri amqp://user:{{ .Values.rabbitmq.auth.password }}@{{ .Release.Name }}-rabbitmq:5672/%2F {{ if .Values.didFinder.site }} --site {{ .Values.didFinder.site }} {{ end }} --threads {{ .Values.didFinder.threads }}"]
```
AndrewEckart: This replaces the `&&` after `sleep 5` with a single `&`. I'm not a bash expert, but won't this cause the `sleep 5` to run in the background? If so I think it would defeat the purpose.

I agree with your comment in #260 that it would be more readable if we could pass a single bash script rather than chaining together a bunch of commands with operators.

AndrewEckart:
Now that I think about it, it's probably better to set the environment variable in the [Dockerfile](https://github.com/ssl-hep/ServiceX-DID-finder/blob/develop/Dockerfile) for the DID finder.

servicex/templates/did-finder/deployment.yaml
ivukotic: here && means continue after this command. One could also use ; instead of &&.
But this code actually has bigger problem: sleep should be in front of the proxy-exporter, so it gives some time to the secret to be created and proxy made. There is no reason to sleep after proxy-exporter.

But anyhow this is all very badly done as the proxy-exporter should be waiting until the x509 proxy appears.

ivukotic:
> Now that I think about it, it's probably better to set the environment variable in the [Dockerfile](https://github.com/ssl-hep/ServiceX-DID-finder/blob/develop/Dockerfile) for the DID finder.

agree... I will change that

ivukotic:
I have reorganized did finder so now it can be started in a clean way

PR #270, opened 2021-03-22
Architecture docs

BenGalewsky:
Thanks @ivukotic for starting this off. Brought in comments from @AndrewEckart and @gordonwatts.

Tried to answer the architecture related questions and cleared up remaining notes. Also reverted an unrelated (and breaking) change to the helm chart


ivukotic (commented):
Thanks for answering some of the questions. 

One small correction:
"Since there may be multiple replcias of each file, the DID finder has an optional
`site` setting which indicates a prefered site from which to serve replicas. 
If there are no replicas hosted at that site, the DID finder will simply return
the first replica from the list returned from Rucio."

_site_ option in Rucio is telling Rucio where is the client, not what is preferred data origin. It uses this information in two ways: a) if the _site_ has xcache, it will prepend all the paths with xcache prefix. b) will try to get replicas listed in order of increasing distance from the _site_.

Also why remove all the remaining questions without addressing them? Even if things were not done it would be nice to know what's the plan, how they will be done.

gordonwatts:
@ivukotic - some of the questions were good for discussion - so they should be pulled out and talked about. But leaving them in an architecture document doesn't make a lot of sense. This document isn't good for question back-and-forths.

PR #275, opened 2021-04-06
adding Release.Name to all the deployments envs

ivukotic:
this is needed so we can add it to logs and later use for aggregations in monitoring.


BenGalewsky:
Did you change the whitespace in these files? The diffs are 100% instead of just your changes. Please make the diffs highlight what is actually changing

ivukotic:
Hi,

This has been fixed.

I

PR #276, opened 2021-04-07
A few changes to fix an error and clarify docs

sthapa:
--namespace argument is in the wrong location and added a bit to the helm chart deploy to emphasize the release name

### DIFF
--- a/docs/deployment/basic.md
+++ b/docs/deployment/basic.md
@@ -39,7 +39,7 @@ pip install servicex-cli
 
 You can run the CLI to install your certs with the following command:
 ```
-servicex init --cert-dir ~/.globus --namespace <default>
+servicex --namespace <default> init --cert-dir ~/.globus 
 ```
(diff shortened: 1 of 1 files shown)


docs/deployment/basic.md
```diff
@@ -124,8 +124,10 @@ helm install -f values.yaml --version v1.0.0-rc.3 servicex ssl-hep/servicex
 ```
 
 Initial deployment is typically rapid, with RabbitMQ requiring up to a minute to
-complete its initialization. After this all the pods of the new deployment 
-should be ready. If you check the status of the pods via
+complete its initialization. The `servicex` argument is used by helm as the release 
+name.  It is used to refer to the chart when deploying, insptacting, or deleting
```
AndrewEckart, line 128:
```suggestion
name.  It is used to refer to the chart when deploying, inspecting, or deleting 
```

AndrewEckart: Looks good. Thanks for the doc fix!

PR #277, opened 2021-04-07
added instance name to the configmap. 

ivukotic:
Getting label from inside app pod is not so easy.


AndrewEckart (approved):
This is a good way to do it.

PR #284, opened 2021-04-30
Updates to the docs for architecture and contributing

sthapa:


AndrewEckart (changes requested):
Thanks! Matching up the microservices to the corresponding repositories is especially helpful.

Please change the label for Preflight Check - it should be the same as the transformers, not ServiceX_App.

sthapa:
Updated the diagram to indicate that pre-flight is just a transformer and has the same repos

PR #287, opened 2021-05-04
Add optional proxyLocation variable for containers running as non-root

holzman:
Some kubernetes deployments (i.e. OpenShift) prefer to run containers
as ephemeral users. This optional variable allows the user to set the
x509 proxy location to somewhere writable by non-root (e.g. /tmp)


BenGalewsky:
Thank you for this, @holzman - It makes it harder to understand the system if there is a default in the docker image that can be overridden here. I would prefer that you move the default to `values.yaml` and always pick it up from the Environment var.

holzman:
Sure, I can do that. (I didn't do it originally since it will break things slightly on upgrade since it adds two new items to values.yaml..)

servicex/values.yaml
```diff
@@ -71,6 +71,7 @@ didFinder:
   rucio_host: https://voatlasrucio-server-prod.cern.ch:443
   auth_host: https://voatlasrucio-auth-prod.cern.ch:443
   threads: 5
+  proxyLocation: /tmp/x509up
```
BenGalewsky, line 74: Should the default be `/etc/grid-security/x509up` to match existing deployments?

holzman: My thought had been to provide a default that worked out of the box for both non-root and root containers, but as long as it's configurable I don't really care.

hollowec:
Another option to solve this problem is to make /etc/grid-security group root writable in the container images, since Openshift/OKD containers run with group root by default.  This is what was done on the ServiceX deployment on OKD at BNL.

holzman:
That's another approach that will work.  My intent was to make things look as familiar as possible (/etc/grid-security not normally writable, proxies live in /tmp)  but I'm OK with just making /etc/grid-security 770.

BenGalewsky:
@holzman Your last comment makes me think we really should bite the bullet and just change all of our images to put the writable proxies in /tmp if that's a normal way of doing things - I'll change everything to make this the hardcoded way and keep config simple.

I appreciate this attempt to not interfere with existing code, but I think this will be better.
Yes?

holzman:
Sounds good to me.

hollowec:
Works for me also.  Thanks!

BenGalewsky:
This PR is obsolete now that we've hardcoded placing user certs in `/tmp` in all of the impacted containers

PR #294, opened 2021-05-20
ServiceX release for POSIX option

zorache:
@oshadura


BenGalewsky:
This looks good! Please update docs/deployment/reference.md

oshadura:
@BenGalewsky we will need for testing in Flux to have released chart with changes (e.g. release candidate)? would it be possible? (The problem is that Flux is deploying Servicex not from sources but from helm release as coffea-casa dependency)

BenGalewsky:
> @BenGalewsky we will need for testing in Flux to have released chart with changes (e.g. release candidate)? would it be possible? (The problem is that Flux is deploying Servicex not from sources but from helm release as coffea-casa dependency)

We can do that with a pre-release tag. I thought that the dev namespace on the UNL cluster allowed `kubectl` directly. In that case you can helm install from a local copy. It would seem painful to develop the chart using flux, but I can accommodate if you wish to go that way.

oshadura:
>>> We can do that with a pre-release tag.

Thanks!!! that would be really great!

About development with Flux it is actually opposite, for coffea-casa we use Flux to deploy from sources (Git repository), so everything is smooth and you saw how it is quick. While for Servicex we deploy it directly from released Helm chart (Helm repository), so basically in this case we provide you kind of 'testing of deployment'. (https://docs.fluxcd.io/projects/helm-operator/en/stable/helmrelease-guide/chart-sources/) @bbockelm should we switch to deploy ServiceX from Git repository?

BenGalewsky:
@oshadura - we should set up @zorache 's deployment of ServiceX specifically to pull from this branch out of GitHub. We can leave the other flux deployments to use the published versions of the chart.

I'm sure Zora will need to make multiple changes to the chart as she goes and so being able to just push to GitHub will be useful

zorache:
(apologies the e57e50b commit was by me and not Oksana)

zorache:
Note: for my testing purposes the namespace configuration for PVC role and rolebinding is currently zora, instead of default.

BenGalewsky:
Moved to PR #315

PR #295, opened 2021-05-20
Add example for accessing flat ntuples

ponyisi:
Notebook with examples from trilepton dataset


BenGalewsky:
@ponyisi - how does this work without the %matplotlib magic at the top? It runs, but no plot shows unless I add that.. Is that a macOS issue, or do you have some external config setting?

ponyisi:
@BenGalewsky - it works for me; I can add %matplotlib at the top, it's a no-op on my system.

BenGalewsky (approved):

PR #311, opened 2021-06-22
Updates based on Andrew and Ilija's feedback

sthapa:


sthapa:
A few updates to the lifecycle and architecture diagrams, discussed with Ilija and Andrew.  Andrew okayed merging this directly.

PR #312, opened 2021-06-22
Update to handle missing secrets

sthapa:
Updated notes to handle situations were secrets aren't given and fix typos


BenGalewsky (approved):
This looks good... is it tied to any existing issues?

Issue #286, opened 2021-05-03
Servicex Helm chart doesn't allow to use of k8s sealed secrets


oshadura:
While adding ServiceX in Flux we found out that Servicex Helm chart doesn't allow to use of k8s sealed secrets:

```
globusClientID:
        valueFrom:
          secretKeyRef:
            name: servicex-globus-auth
            key: globusClientID
```
 and other cases.

cc: @jthiltges 

The Servicex uses an envvar to find the config file, but the chart doesn't let you use your own config. https://github.com/ssl-hep/ServiceX/blob/develop/servicex/templates/app/deployment.yaml#L21-L22

And at the same time, ServiceX application itself only looks at the config file, no envvars. https://github.com/ssl-hep/ServiceX_App/blob/0c9df3029b9e1e157416e239454698a4a0419bc1/servicex/web/utils.py#L5-L7

Trying to use globusClientID from secrets at https://cmsaf-servicex.servicex.coffea.casa/ fails with `client_id must be a valid UUID.`, while Globus UUID is correct.

AndrewEckart:
Sounds like the chart should support setting environment variables, and the API server should read its configuration from environment variables, not just a configuration file.

Both parts should be doable. The second part requires us to decide whether environment variables should take precedence over the config file or vice versa. @BenGalewsky, what do you think? This [SO post](https://stackoverflow.com/questions/11077223/what-order-of-reading-configuration-values) suggests environment variables should overrule config files.

AndrewEckart:
Looking around for best practices on injecting environment variables into templates. The PostgreSQL chart is a good example:


extraEnv | Any extra environment variables you would like to pass on to the pod. The value is evaluated as a template. | []
-- | -- | --
extraEnvVarsCM | Name of a Config Map containing extra environment variables you would like to pass on to the pod. The value is evaluated as a template. | nil

[Template code](https://github.com/bitnami/charts/blob/1dbd2a203831610986e4998f34e8149cc069bf9b/bitnami/postgresql/templates/statefulset.yaml#L239-L241) - note the reliance on `common.tplvalues.render` from the [Bitnami Common](https://artifacthub.io/packages/helm/bitnami/common) library chart. We would have to add it as a dependency, but it's well-maintained.

BenGalewsky:
Could we put the entire config file into a sealed secret and mount it instead of the configMap?

AndrewEckart:
[...]

PR #322, opened 2021-07-19
Mount secrets into App using env vars

BenGalewsky:
# Problem
The helm chart doesn't support sealed secrets to secure confidential app parameters.

Partial solution to #286 

# Approach
User provides the name of an unsealed secret in a new value: `secrets` - if this is set, then the environment variables are set in the app pod as part of the deployment spec.

This PR includes a template for what a secret file should look like and instructions on installing and using the `SealedSecrets` operator.


sthapa:
LGTM

PR #344, opened 2021-09-09
Updates to support secrets

sthapa:
Update deployments for did finders and preflight to get rabbitmq pass from secrets

### DIFF
--- a/example_secrets.yaml
+++ b/example_secrets.yaml
@@ -10,13 +10,11 @@ data:
   globusClientSecret: <<secret key>>
   secretKey: <<secret key for app>>
   jwtSecretKey: <<jwtSecretKey>>
-  minioAccessKey: <<minioAccessKey>>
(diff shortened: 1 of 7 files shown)


AndrewEckart (approved):
LGTM!

Suggest versioning up to chart version 1.0.14.

Not sure what will happen if certain keys are missing from the secrets. Ideally Helm would give an error, but this may lead to errors at runtime.

sthapa:
> LGTM!
> 
> Suggest versioning up to chart version 1.0.14.
> 
> Not sure what will happen if certain keys are missing from the secrets. Ideally Helm would give an error, but this may lead to errors at runtime.

You'll get an error when the pods try to startup and are missing secrets.

BenGalewsky (approved):
Thanks for figuring out how to get the Postgres and Rabbit secrets into the various config settings!

PR #345, opened 2021-09-10
Add ignore for *~

sthapa:


AndrewEckart (approved):
I usually stick to IDEs over Emacs/Vim/Gedit, but I don't see this conflicting with any other file names or extensions.

PR #348, opened 2021-09-21
Env value expects a quoted string

BenGalewsky:
Sigh, I should have tested this before merging...

# Problem 
When deploying the rucio did finder you get this error from helm:
```
Error: UPGRADE FAILED: failed to create resource: Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found 4, error found in #10 byte of ...|,"value":41.78},{"na|..., bigger context ...|rabbitmq:5672"},{"name":"RUCIO_LATITUDE","value":41.78},{"name":"RUCIO_LONGITUDE","value":-87.7},{"n|...
```


ivukotic (approved):
LGTM.

PR #351, opened 2021-09-24
Secret name fix

sthapa:
This fixes an issue where the chart uses secretkey and secretKey to do too different things.  secretkey is used by the minio chart and secretKey is used by the web app to for flask.  secretKey is changed to use flaskSecretKey in order to disambiguate the two.


BenGalewsky (approved):
Ouch - that must have been hard to track down!
Thank you

PR #353, opened 2021-09-24
add priorityClassName config to transfomer deployment

fengpinghu:
This is to enable passing priorityClassName to the transformer pods.

### DIFF
--- a/servicex/templates/app/configmap.yaml
+++ b/servicex/templates/app/configmap.yaml
@@ -72,6 +72,10 @@ data:
 
     TRANSFORMER_NAMESPACE="{{ .Release.Namespace }}"
 
+    {{ if .Values.transformer.priorityClassName }}
+    TRANSFORMER_PRIORITY_CLASS="{{ .Values.transformer.priorityClassName }}"
+    {{ end }}
+
     TRANSFORMER_PULL_POLICY = '{{ .Values.transformer.pullPolicy }}'
 
     TRANSFORMER_MANAGER_ENABLED = True
--- a/servicex/values.yaml
+++ b/servicex/values.yaml
@@ -112,6 +112,9 @@ transformer:
     minReplicas: 1
     maxReplicas: 20
   pullPolicy: Always
+  # Add priorityClassName to control priority and preemption of the transformer
+  # pods, Priorityclass needs to be created in the cluster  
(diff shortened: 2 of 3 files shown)


servicex/templates/app/configmap.yaml
```diff
@@ -72,6 +72,10 @@ data:
 
     TRANSFORMER_NAMESPACE="{{ .Release.Namespace }}"
 
+    {{ if .Values.transformer.priorityClassName }}
+    TRANSFORMER_PRIORITY_CLASS="{{ .Values.transformer.priorityClassName }}"
```
BenGalewsky, line 76: A default needs to be added to Servicex/values.yaml and documented in docs/deployment/reference.md

BenGalewsky:
We want to make sure that this easily deploys on any cluster. It seems like  PriorityClasses are not standard... should we add creation of one to the helm chart? It looks like `system-cluster-critical` and `system-node-critical` are always available, but maybe not a good one to use

fengpinghu:
Would it make sense to not create a default. The transformer deployment can
work perfectly fine if user don't supply the PriorityClass. They may just
get a global default from the cluster it has that. They will run ok too
without this setting.
In our case on the AF cluster we have a PriorityClass called normal which
is the global default. This priorityclass won't preempt other pods.
So without prorityclass setting,  the transformer pods will get the normal
priorityclass and will not preempt other pods.  But say we like
the transformer pods to be able to preempt another big resource consumer --
the htcondor pods, we can just pass a value to achieve this goal but leave
the global default unchanged.

servicex/values.yaml
```diff
@@ -112,6 +112,9 @@ transformer:
     minReplicas: 1
     maxReplicas: 20
   pullPolicy: Always
+  # Add priorityClassName to control priority and preemption of the transformer
+  # pods, Priorityclass needs to be created in the cluster  
+  # priorityClassName:
```
BenGalewsky: Actually, this works if you just leave a blank after the colon:

```yaml
  pullPolicy: Always
  priorityClassName:
```
It would be a little cleaner that way instead of commented out lines.

BenGalewsky (approved):
Great - I look forward to seeing what this looks like in practice on a cluster that supports priority class

PR #354, opened 2021-09-28
removed threads as it won't be needed with the new did finder

ivukotic:
simple cleanup. merge only after https://github.com/ssl-hep/ServiceX_DID_Finder_Rucio/pull/41


BenGalewsky (approved):
Thanks for tidying this up

PR #355, opened 2021-09-29
Add support for adding a prefix

sthapa:
Sets env variables in did finder containers to use a cache prefix

### DIFF
--- a/servicex/templates/did-finder-cernopendata/deployment.yaml
+++ b/servicex/templates/did-finder-cernopendata/deployment.yaml
@@ -21,6 +21,10 @@ spec:
         env:
           - name: INSTANCE_NAME
             value: {{ .Release.Name }}
+        {{- if .Values.didFinder.CERNOpenData.cachePrefix }} 
+          - name: CACHE_PREFIX
+            value: "{{ .Values.didFinder.CERNOpenData.cachePrefix }}"
+        {{- end }}
         {{- if .Values.secrets }}
           - name: RMQ_PASS 
             valueFrom:
--- a/servicex/templates/did-finder-rucio/deployment.yaml
+++ b/servicex/templates/did-finder-rucio/deployment.yaml
@@ -35,6 +35,10 @@ spec:
           - name: RMQ_URI
             value: amqp://user:{{ .Values.rabbitmq.auth.password }}@{{ .Release.Name }}-rabbitmq:5672
     {{- end }}
+    {{- if .Values.didFinder.rucio.cachePrefix }}
+          - name: CACHE_PREFIX
+            value: "{{ .Values.didFinder.rucio.cachePrefix }}"
+    {{- end }}
           - name: RUCIO_LATITUDE
             value: "{{ .Values.didFinder.rucio.servicex_latitude }}"
           - name: RUCIO_LONGITUDE
--- a/docs/deployment/reference.md
+++ b/docs/deployment/reference.md
@@ -39,6 +39,7 @@ parameters for the [rabbitMQ](https://github.com/bitnami/charts/tree/master/bitn
 | `didFinder.rucio.pullPolicy`         | Rucio DID Finder image pull policy                     | `Always`                                          |
 | `didFinder.rucio.servicex_latitude`        | Latitude of the computing center where ServiceX runs. Will be used by Rucio to return the closest input data replica.        | 41.78 |
 | `didFinder.rucio.servicex_longitude`        | Longitude of the computing center where ServiceX runs. Will be used by Rucio to return the closest input data replica.        | -87.7 |
+| `didFinder.rucio.cachePrefix`        | Prefix string to stick in front of file paths. Useful for XCache | |
 | `didFinder.rucio.rucio_host`         | URL for Rucio service to use                     | `https://voatlasrucio-server-prod.cern.ch:443`          |
 | `didFinder.rucio.auth _host`         | URL to obtain Rucio authentication               | `https://voatlasrucio-auth-prod.cern.ch:443`            |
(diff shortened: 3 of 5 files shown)


servicex/templates/did-finder-cernopendata/deployment.yaml
```diff
@@ -21,6 +21,10 @@ spec:
         env:
           - name: INSTANCE_NAME
             value: {{ .Release.Name }}
+        {{- if .Values.didFinder.CERNOpenData.cachePrefix }}
```
BenGalewsky, line 24: Is there any support for this in the CERNOpenData DID Finder? If not let's not confuse people by adding options that don't do anything. We can add this when the DID finder supports it

servicex/templates/did-finder-rucio/deployment.yaml
```diff
@@ -34,6 +34,10 @@ spec:
     {{- else }}
           - name: RMQ_URI
             value: amqp://user:{{ .Values.rabbitmq.auth.password }}@{{ .Release.Name }}-rabbitmq:5672
+    {{- end }}
+    {{- if .Values.didFinder.rucio.cachePrefix }}
+          - name: CACHE_PREFIX
```
BenGalewsky, line 39: The Rucio DID Finder uses a [command line arg](https://github.com/ssl-hep/ServiceX_DID_Finder_Rucio/blob/519b3f52ce338870c8a3f34498947b5b592a1a74/scripts/did_finder.py#L46) for this, not an env var... this should go into the command

gordonwatts:
WAIT! Now I understand this.

Isn't this for adding a prefix to a root:// file? Shouldnt' this just be done in the library so every DID finder that returns a root:// file can add a prefix and have it locally cached?

BenGalewsky:
> Isn't this for adding a prefix to a root:// file? Shouldnt' this just be done in the library so every DID finder that returns a root:// file can add a prefix and have it locally cached?

Can we think of any other non Rucio DID finders that would return root:// files?

servicex/templates/did-finder-rucio/deployment.yaml
sthapa: The container calls a runme.sh that then calls the did finder script with the RMQ_URI command line.  This just follows that pattern and avoids having to parse args in a bash script.

servicex/templates/did-finder-cernopendata/deployment.yaml
sthapa: @ivukotic  indicated that the CernOpenData finder supports this, I'm planning on adding support there in consultation with @gordonwatts once the rucio finder support has landed (Ilija and I are testing that right now)

sthapa:
@gordonwatts I considered modifying the library but it doesn't look like we have the paths until the LookupRequest call returns.  So  I don't think we can add the prefix support to the library.  

@BenGalewsky @ivukotic indicated that the CernOpenData could use xcache or squid as a proxy.  I think if the finder returns root URIs then we can prefix it, for http urls, we'd probably need to set HTTP_PROXY env var in the transformer container and hope the code downloading the files understands that and uses it.

gordonwatts:
> Can we think of any other non Rucio DID finders that would return root:// files?

The CERN DID finder, and anyone else that wants to store things via the root protocol. So - potentially a bunch in HEP since it is the defacto data distribution protocol here.

gordonwatts:
> @gordonwatts I considered modifying the library but it doesn't look like we have the paths until the LookupRequest call returns. So I don't think we can add the prefix support to the library.

I'm not sure I understand:

* To get Rabbit MQ stuff, the library has to parse the command line arguments
* Just add the prefix as another command line argument

You can see [this code here](https://github.com/ssl-hep/ServiceX_DID_Finder_Rucio/blob/519b3f52ce338870c8a3f34498947b5b592a1a74/scripts/did_finder.py#L49) where the library gets to add the command line arguments, and [here where it calls back into the library](https://github.com/ssl-hep/ServiceX_DID_Finder_Rucio/blob/519b3f52ce338870c8a3f34498947b5b592a1a74/scripts/did_finder.py#L77) with the arguments parsed.

Logically you should not have to touch the DID finder or CERN Open Data other than to modify the version of the library they are using if you use this approach. In fact, this code pattern was designed to support exactly this.

Another obvious thing to support is caching of queries - so if two queries come in that are the same, why even invoke the thing? The library is built so that it could be extended to support all that.

In the actual library you'd need to:

* Add the correct flags (somewhere near [here](https://github.com/ssl-hep/ServiceX_DID_Finder_lib/blob/e5021a4d22c45f524d5615c3a5054e6e8cc0f25d/src/servicex_did_finder_lib/communication.py#L140))
* Modify the return logic to look for the `root://` files coming back and add the prefix (some return an `http://` interface) (somewhere near [here](https://github.com/ssl-hep/ServiceX_DID_Finder_lib/blob/e5021a4d22c45f524d5615c3a5054e6e8cc0f25d/src/servicex_did_finder_lib/communication.py#L36))
* Add some tests to make sure this works! :-)

The only files it would then miss are when the services app gets a list of addresses to access. That avoids a did finder in the current logic (maybe it shouldn't?). I'd say fixing that was out of scope for this planned work.

sthapa:
@gordonwatts okay, I'll look at the library and add the changes there once I start adding support for prefixes to the CERNOpenData finder.  With @ivukotic's metalink changes to the rucio finder, I had to take a slightly different approach since he architected the code a bit differently. I'll submit a PR to use the library to set the argument shortly.

BenGalewsky (approved):
@sthapa Any chart changes that result in new values should include a line in `values.yaml` as well as an entry in the reference docs. I pushed a commit to this branch to add those so we can get this PR deployed

PR #373, opened 2021-11-23
Add external minio support

sthapa:
I checked the servicex app code and it looks like this should work but I'm doing a quick test that should confirm this later this afternoon.

### DIFF
--- a/servicex/templates/app/configmap.yaml
+++ b/servicex/templates/app/configmap.yaml
@@ -103,9 +103,9 @@ data:
 
     {{ if .Values.objectStore.enabled }}
     OBJECT_STORE_ENABLED = True
-    {{ if  .Values.minio.publicurl }}
-    MINIO_URL = '{{ .Values.minio.publicurl }}'
-    MINIO_TRANSFORMER = '{{ .Values.minio.publicurl }}'
+    {{ if not .Values.objectStore.internal }}
+    MINIO_URL = '{{ .Values.objectStore.publicURL }}'
+    MINIO_TRANSFORMER = '{{ .Values.objectStore.publicURL }}'
     {{ else }}
     MINIO_URL = '{{ .Release.Name }}-minio:{{ .Values.minio.service.port }}'
     MINIO_URL_TRANSFORMER = '{{ .Release.Name }}-minio:{{ .Values.minio.service.port }}'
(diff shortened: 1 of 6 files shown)


BenGalewsky:
Not quite so easy - you need to prevent the helm chart from deploying a minio instance

sthapa:
@BenGalewsky this now disables the minio pod deploy and should set the various settings.  Mind taking a quick glance at it while I test?

servicex/templates/app/configmap.yaml
```diff
@@ -103,23 +103,32 @@ data:
 
     {{ if .Values.objectStore.enabled }}
     OBJECT_STORE_ENABLED = True
+      {{ if not .Values.objectStore.internal }}
+    # using external minio 
+    MINIO_URL = '{{ .Values.objectStore.publicURL }}'
+    MINIO_URL_TRANSFORMER = '{{ .Values.objectStore.publicURL }}'
+    MINIO_PUBLIC_URL = '{{ .Values.objectStore.publicURL }}'
+    MINIO_SECURED = {{ ternary "True" "False" .Values.objectStore.usetls }}
```
BenGalewsky, line 111: This is eventually a python config file. You should remove the quotes around `True` and `False`

BenGalewsky (changes requested):
Please update `docs/deployment/reference.md` to show the new settings

servicex/templates/app/configmap.yaml
sthapa: Without the quotes, helm looks for a function called True and errors out.  The configmap gets the strings without the quotes.

sthapa: Also updated the docs with the new settings

BenGalewsky: I don't see any changes to the [deployment reference docs](https://github.com/ssl-hep/ServiceX/blob/b28498a96a5df99f57ac0a261ea679175f02cf70/docs/deployment/reference.md?plain=1#L8)

servicex/values.yaml
```diff
@@ -10,6 +10,7 @@ secrets:
 # want to have the option of delivering results as parquet files
 objectStore:
   enabled: true
+  internal: false
   publicURL:
```
BenGalewsky, line 15: This should default to `true` to keep existing behavior. It will break all current deployments otherwise

BenGalewsky: Pushed a fix

servicex/templates/app/configmap.yaml
BenGalewsky: Pushed an update

BenGalewsky (approved):

Issue #369, opened 2021-11-19
Rucio DID Finder Returns Logical Filenames with Prefix


BenGalewsky:
# Problem
In _coffea-casa_ the xCache is configured to use a "redirector" which does not want the physical file paths, but only needs the logical files.

# Approach
Add an option to the Rucio DID Finder to return logical file paths and skip the `file-replica` lookup.

i.e.

Don't return `root://fax.mwt2.org:1094//pnfs/uchicago.edu/atlaslocalgroupdisk/rucio/user/kchoi/4d/ad/ttHML_80fb_410220_mc16d.root`

instead, returns `user.kchoi:ttHML_80fb_410220_mc16d.root`

oshadura:
Just to show here an example: 
```
[root@cmsaf-servicex-did-finder-rucio-5fc4999547-h9td8 app]# rucio list-files cms:DYJetsToLL_M-50_TuneCP5_13TeV-amcatnloFXFX-pythia8/RunIIAutumn18NanoAODv7-Nano02Apr2020_102X_upgrade2018_realistic_v21_ext2-v1/NANOAODSIM
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------+-------------+------------+----------+
| SCOPE:NAME                                                                                                                                                                                               | GUID   | ADLER32     | FILESIZE   | EVENTS   |
|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------+-------------+------------+----------|
| cms:/store/mc/RunIIAutumn18NanoAODv7/DYJetsToLL_M-50_TuneCP5_13TeV-amcatnloFXFX-pythia8/NANOAODSIM/Nano02Apr2020_102X_upgrade2018_realistic_v21_ext2-v1/70000/553F8669-95F5-9940-82FE-602AC2423121.root  | (None) | ad:60e81f96 | 426.931 MB |          |
| cms:/store/mc/RunIIAutumn18NanoAODv7/DYJetsToLL_M-50_TuneCP5_13TeV-amcatnloFXFX-pythia8/NANOAODSIM/Nano02Apr2020_102X_upgrade2018_realistic_v21_ext2-v1/130000/244509DB-FF94-9349-BB17-CF7036ABED55.root | (None) | ad:62f9d590 | 1.472 GB   |          |
[...]

PR #374, opened 2021-11-24
Add option to tell rucio DID Finder to report logical names

BenGalewsky:
# Problem
The CMS deployment of xCache assumes that the URIs for Root files is the logical name (not a replica URL)

Partial solution to [ServiceX issue 369](https://github.com/ssl-hep/ServiceX/issues/369)

# Approach
Added new value: `didFinder.rucio.reportLogicalFiles` 
 
If this is set to true, it passes an environment var down to the Rucio DID Finder telling it to return the `identity` property instead of the `url`


ivukotic (approved):
ah now I see. I should have reviewed this one first. LGTM

PR #375, opened 2021-11-24
Change api to allow running on kubernetes 1.22

sthapa:
We are using the deprecated extensions/v1beta1 for ingress config, that's been dropped as of [k8s 1.22](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#:~:text=The%20extensions%2Fv1beta1%2C%20apps%2F,9.).  Update the api version and make changes to  use the new api.  

The only thing that isn't quite straightforward is the type (i.e. using Prefix vs ImplementationSpecific).  I think Prefix is what we want.

### DIFF
--- a/servicex/requirements.yaml
+++ b/servicex/requirements.yaml
@@ -4,7 +4,7 @@ dependencies:
     repository: https://charts.bitnami.com/bitnami/
   - name: minio
     version: ">=9.0.0"
-    repository: https://helm.min.io/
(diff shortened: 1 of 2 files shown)


sthapa:
Did some testing with a local minikube instance and the ingress controller/functionality appears to be working with the changes.

BenGalewsky (approved):
Nice! No more warnings

PR #376, opened 2022-01-05
Make sure rabbitmq liveness probe is enabled

sthapa:
This makes sure that the liveness probe is enabled and running for rabbitmq.  Rabbitmq crashing but sticking around has caused at least 2 problems on instances.  Although the liveness probe should be running based on the chart, this makes that explicit.


ivukotic (approved):
LGTM

Issue #337, opened 2021-09-01
Remove Preflight


BenGalewsky:
The preflight check has been disabled for quite some time. While it's possible that there are some one-off checks that could be done before launching transformers we currently don't have any.

Part of Epic #309 

Allowing multiple code generators in a deployment creates some challenges in continuing to support the pre-flight check without impacting round-trip time for smaller transforms.

In the spirit of [YAGNI](https://martinfowler.com/bliki/Yagni.html) we will remove this capability until such time that we find a good use for it.

Tasks:

- [ ] Remove preflight from workflow. DID Finder no longer uses `/servicex/internal/transformation/<string:request_id>/preflight` endpoint, but directly invokes the `/servicex/internal/transformation/<string:request_id>/start` endpoint
- [ ] Remove unused endpoints from app 
- [ ] Remove preflight deployment from helm chart 
- [ ] Delete preflight scripts from transformer images

PR #378, opened 2022-01-25
preflight removal

ivukotic:
since it is not needed

Partial solution to [ServiceX Issue 337](https://github.com/ssl-hep/ServiceX/issues/337)


BenGalewsky (approved):
Yes, this gets it all

PR #379, opened 2022-02-01
Minio https fixes

sthapa:
Update charts to allow admins to specify using https connections for client connections and simplify logic for server https connections

### DIFF
--- a/servicex/templates/app/configmap.yaml
+++ b/servicex/templates/app/configmap.yaml
@@ -115,7 +115,7 @@ data:
     MINIO_URL = '{{ .Values.objectStore.publicURL }}'
     MINIO_URL_TRANSFORMER = '{{ .Values.objectStore.publicURL }}'
     MINIO_PUBLIC_URL = '{{ .Values.objectStore.publicURL }}'
-    MINIO_SECURED = {{ ternary "True" "False" .Values.objectStore.usetls }}
+    MINIO_SECURED = {{ ternary "True" "False" .Values.objectStore.useTLS }}
       {{ else }}
     # using internal minio 
     MINIO_URL = '{{ .Release.Name }}-minio:{{ .Values.minio.service.port }}'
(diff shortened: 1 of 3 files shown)


servicex/templates/app/configmap.yaml
```diff
@@ -103,12 +103,19 @@ data:
 
     {{ if .Values.objectStore.enabled }}
     OBJECT_STORE_ENABLED = True
+    # default to using a https connection to the client unless explicitly
+    # turned off
+      {{ if .Values.objectStore.useClientTLS  }}
+    MINIO_CLIENT_SECURED = True
```
BenGalewsky: Maybe instead of `MINIO_CLIENT_SECURED ` and `MINIO_SECURED` it would be clearer to be `MINIO_INTERNAL_SECURED` and `MINIO_EXTERNAL_SECURED` This is such a confusing topic we need all the clarity we can muster!

servicex/values.yaml
```diff
@@ -13,6 +13,12 @@ objectStore:
   # Enable deployment of Minio Object Store with this chart
   internal: true
   publicURL:
+  # turn off https connections to minio, should be off if using internal
+  # minio
+  useTLS: false
```
BenGalewsky, line 18: Likewise maybe `useTLSInternally` and `useTLSExternally`

servicex/templates/app/configmap.yaml
sthapa: The tricky part is that the differentiation here is on the client connections vs the server to minio connections.  `MINIO_SECURED` can determine whether connections within the k8s cluster is encrypted or it can determine whether connections from the k8s cluster to another site is encrypted.  So using `MINIO_INTERNAL_SECURED` and `MINIO_EXTERNAL_SECURED` might confuse admins as to how to encrypt a connection from ServiceX server components in a K8S cluster to an external minio instance living somewhere else or living outside the K8S cluster.

sthapa:
Update param names

BenGalewsky (approved):
There was a typo in `values.yaml` that prevented this from working correctly for public encrypted minio.

I fixed this and tested it with a local development deploy as well as a remote deploy. I haven't tested with an external object store, but it looks like it should work.

PR #382, opened 2022-02-17
Remove hard-coded images from templates

holzman:
We tend to hit dockerhub pull limits, so it's good to be able to customize all images to point at a local cache.


BenGalewsky:
This makes sense -- I'll merge it once we get reliable integration tests running on dev (also we are also trying to keep the U of C deployments stable for a high profile demo next week

PR #387, opened 2022-02-18
Add server status links

BenGalewsky:
# Problem
Users need to know that the production ServiceX instances are up and running

Solves #383 

# Approach
Add links to the GitActionBoard from the README and the _Getting Started_ guide.


sthapa (approved):
lgtm

Issue #385, opened 2022-02-18
Include Helm Chart in CI Process


BenGalewsky:
It doesn't look like Helm charts can use calver, but we need to create a tagged version and update the `values.yaml` file to point to the versioned tags

PR #388, opened 2022-02-22
tag_and_release script

BenGalewsky:
# Problem
There are several steps to create a new helm chart release that has docker images pinned to a tag. We want to do this often.

Solution to issue #385 

# Approach
Create a helm chart release pinned to tagged serviceX docker images. The approach taken is that the chart on the `develop` branch will always point to the released docker images. If you want to test out some of the development images you have to explicitly set this in your personal `values.yaml`.

This script will update `values.yaml` to point to the released tags, push to develop, tag `develop` branch and publish the helm chart.

### DIFF
--- a/tag_and_release.sh
+++ b/tag_and_release.sh
@@ -31,6 +31,14 @@
 #
 if [ "$#" -ne 2 ]; then
     echo "Usage: tag_release.sh serviceX_version chart_version"
+    exit 1
+fi
+
+if [ ! -d "../ssl-helm-charts" ]; then
+  echo "Helm chart repo not available. Creating one now"
+  pushd ..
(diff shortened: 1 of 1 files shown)


sthapa:
Two things I noticed, the usage check at the top should exit.  Right now, if you don't give the right number of arguments, it prints the usage and then continues with the rest of the code.

The other thing, is the code seems to assume that there's a ssl-helm-charts directory present.  If it's not there, line 59 moves and renames  the tarball.  I couldn't see a check or anything in the prior lines that creates that directory.

sthapa (approved):
LGTM, although it might be good to add a set -e at the top of the file to exit on error or to add checking to exit if any of the git or helm commands fail.  Not really necessary though.

PR #389, opened 2022-03-04
update logstash version

ivukotic:


sthapa:
The changes look good, however there appears to be a typo in one of the file names.  I think it should be daemonset.yaml and not deamonset.yaml.  Is this going to affect deployment or usage?

ivukotic:
> 

This is just a helm thing, one can make whatever filename you want.
I will still correct it.

sthapa (approved):
LGTM

PR #391, opened 2022-03-11
Adding about

ivukotic:
just adding a few variables to the config map in order to get about page values.

### DIFF
--- a/servicex/templates/app/configmap.yaml
+++ b/servicex/templates/app/configmap.yaml
@@ -14,18 +14,6 @@ data:
     CHART = '{{ .Chart.Name }}-{{ .Chart.Version }}'
     APP_IMAGE_TAG = '{{ .Values.app.tag }}'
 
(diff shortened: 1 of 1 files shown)


servicex/templates/app/configmap.yaml
```diff
@@ -91,7 +94,6 @@ data:
     {{- else }}
     TRANSFORMER_X509_SECRET=None
     {{- end }}
-    TRANSFORMER_VALIDATE_DOCKER_IMAGE = {{- ternary "True" "False" .Values.app.validateTransformerImage }}
```
BenGalewsky, line 94: Why would you delete this? - This is a setting that states whether to validate the provided transformer image exists on DockerHub. We want this in prod, but not in development

ivukotic: Ah I thought this was preflight image... I will put it back.

BenGalewsky (approved):
Clever solution to this problem!

PR #397, opened 2022-03-21
Split default Transformer image name and tag into two parameters

ivukotic:
Without this it is hard to get correct transformer image tag.

### DIFF
--- a/docs/deployment/reference.md
+++ b/docs/deployment/reference.md
@@ -1,7 +1,7 @@
-# ServiceX Helm Chart Reference 
+# ServiceX Helm Chart Reference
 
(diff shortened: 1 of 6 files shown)


gordonwatts:
A little more info:

* We discovered this because @ivukotic got pissed off at the failed test email from the backend tests (so the infrastructure is working!)
* The issue is that the transformer image is set by each deployment (xaod, cms, uproot), but you want the tag set by the services release
* Since the current version has them common, the image tag is getting lost.
* This leads to ServiceX trying to run a transformer by default with no tag, and bad-things-happen.

All the work, as you can see, was done by @ivukotic to fix this. I'm just adding some context.

BenGalewsky (approved):

PR #400, opened 2022-03-22
Update to use newer bitnami/minio chart

holzman:


sthapa:
We'll need to make sure that we make sure that any sites using this set the rootUser and rootAccount properly.  Otherwise an upgrade from the 8.0.10 will break or worse result in a non-functioning install.  Two comments, can you limit the versions for minio to 11.1.* or something similar?  Saying we'll accept anything newer than 11.0 might cause problems if there's a future version of the chart that's incompatible.  Also, it looks like your changes for the ncsa and alpine locations snuck into this PR.

holzman:
Thanks for the review.

- I based this off of my "remove hard-coded images" branch since I figured that would be merged first anyhow.
- Locking in a specific major/minor version of minio makes sense.
- I'll add some code to throw an error if `rootUser` and `rootPassword` aren't set.

I'll also rebase off of main instead of the feature branch if it isn't merged by the time I push.

sthapa:
I don't think you need code checks for `rootUser` and `rootPassword` the bitnami chart already outputs an error and helm will abort the chart upgrade.  

```
PASSWORDS ERROR: You must provide your current passwords when upgrading the release.
Note that even after reinstallation, old credentials may be needed as they may be kept in persistent volume claims.
Further information can be obtained at https://docs.bitnami.com/general/how-to/troubleshoot-helm-chart-issues/#credential-errors-while-upgrading-chart-releases
```

I don't think you need to rebase, we can sequence the PR merges so that this happens after #401 . 

My main concern with the config parameter change is documenting it so that admins will be able make the needed changes in order to update successfully.  E.g. anyone using flux will need to update their secrets to include the new keys and where the secret is specified.

holzman:
> My main concern with the config parameter change is documenting it so that admins will be able make the needed changes in order to update successfully. E.g. anyone using flux will need to update their secrets to include the new keys and where the secret is specified.

Sure, makes sense. Where do you want me to document this change in behavior? 

I think we probably want minio updated before the next workshop (I think there are 4 CVEs open on the version that's being packaged now).

sthapa:
I'll write up docs for this today

oshadura:
We cant deploy ServiceX on new UNL cluster running k8s 1.23, because of old minio and rabbitmq version both looks like using networking.k8s.io/v1beta1
```
{"level":"error","ts":"2022-04-15T15:39:15.085Z","logger":"controller.helmrelease","msg":"Reconciler error","reconciler group":"helm.toolkit.fluxcd.io","reconciler kind":"HelmRelease","name":"opendataaf-servicex","namespace":"opendataaf-dev","error":"Helm install failed: unable to build kubernetes objects from release manifest: unable to recognize \"\": no matches for kind \"Ingress\" in version \"networking.k8s.io/v1beta1\""}
```
cc @jthiltges

holzman:
Rebased against develop.

BenGalewsky:
I would like to merge this to develop and request @sthapa, @oshadura show us the specifics of how the flux delployments have to be updated to work with this new sub chart

sthapa (approved):
Changes look good, I'll add upgrade instructions for admins

PR #401, opened 2022-03-22
Remove hard-coded images from templates

holzman:
This is reopening https://github.com/ssl-hep/ServiceX/pull/382 on a feature branch.

### DIFF
--- a/docs/deployment/reference.md
+++ b/docs/deployment/reference.md
@@ -10,6 +10,7 @@ parameters for the [rabbitMQ](https://github.com/bitnami/charts/tree/master/bitn
 | `app.image`                          | ServiceX_App image name                          | `sslhep/servicex_app`                                   |
 | `app.tag`                            | ServiceX image tag                               | `latest`                                                |
 | `app.pullPolicy`                     | ServiceX image pull policy                       | `Always`                                          |
+| `app.checksImage`                    | ServiceX init container image for checks         | `ncsa/checks:latest`                                    |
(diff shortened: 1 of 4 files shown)


sthapa:
This looks good but let me get this tested today

sthapa:
After some thought, can you change the chart to use a if then so that if the image isn't specified, it'll automatically use the docker images but will use the user specified image otherwise?  I think that'll minimize the disruption and let us set a default in case we want to switch to use harbor or some other source.

holzman:
That's why I also changed values.yaml - to provide those defaults (I thought that was best practice).

Do you really want the default values hard-coded into the template?

BenGalewsky:
This is so odd, I could have sworn I merged this months ago. Still...

BenGalewsky (changes requested):
Please update `docs/deployment/reference.md` to show these new values

holzman:
Sure.  Do you want the rest of the commit as-is, or did you want the default in the templates as @sthapa suggested?

BenGalewsky:
Yes - the rest of the commit is perfect. That indeed is what defaults in `values.yaml` is for!

sthapa:
I missed the values.yaml changes but with that, it should be fine.  No need to add conditional logic for this

holzman:
Done.

holzman:
Any issues with merging this? It's been sitting for 3 weeks.

BenGalewsky (approved):

Issue #399, opened 2022-03-21
With 20220311-2056-stable and enabled Posix persistence volume support, services app pod is keeping on crashing


oshadura:
**Describe the bug**
With 20220311-2056-stable and enabled Posix persistence volume support, services app pod is keeping on crashing

**To Reproduce**
Adding in values.yaml: 
```
objectStore:
      enabled: false
transformer:
      persistence:
        existingClaim: skyhook-pv-claim
        subdir: /mnt/cephfs/
```
I see that opendataaf-servicex-servicex-app-6d9c6569c5-8nkmh   is keeping on crashing with  `CrashLoopBackOff`:
```
[2022-03-21 18:44:38 +0000] [20] [ERROR] Exception in worker process
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/gunicorn/arbiter.py", line 589, in spawn_worker
    worker.init_process()
  File "/usr/local/lib/python3.6/site-packages/gunicorn/workers/base.py", line 134, in init_process
    self.load_wsgi()
  File "/usr/local/lib/python3.6/site-packages/gunicorn/workers/base.py", line 146, in load_wsgi
    self.wsgi = self.app.wsgi()
  File "/usr/local/lib/python3.6/site-packages/gunicorn/app/base.py", line 67, in wsgi
    self.callable = self.load()
  File "/usr/local/lib/python3.6/site-packages/gunicorn/app/wsgiapp.py", line 58, in load
    return self.load_wsgiapp()
  File "/usr/local/lib/python3.6/site-packages/gunicorn/app/wsgiapp.py", line 48, in load_wsgiapp
    return util.import_app(self.app_uri)
  File "/usr/local/lib/python3.6/site-packages/gunicorn/util.py", line 412, in import_app
    app = app(*args, **kwargs)
  File "/home/servicex/servicex/__init__.py", line 141, in create_app
    app.config['TRANSFORMER_NAMESPACE']):
  File "/home/servicex/servicex/transformer_manager.py", line 218, in persistent_volume_claim_exists
    pvcs = api.list_namespaced_persistent_volume_claim(namespace=namespace, watch=False)
  File "/usr/local/lib/python3.6/site-packages/kubernetes/client/api/core_v1_api.py", line 15533, in list_namespaced_persistent_volume_claim
    return self.list_namespaced_persistent_volume_claim_with_http_info(namespace, **kwargs)  # noqa: E501
[...]

PR #402, opened 2022-03-24
Add get access on persistentvolumeclaims to ServiceX service account

BenGalewsky:
# Problem
On clusters with real security (as compared to DockerDesktop) - The serviceX app fails when configured for posix volumes for transformer results.
 
Solves #399 

# Approach
Add the list permission on PVCs for the ServiceX service account


sthapa (approved):
LGTM

PR #403, opened 2022-03-28
Allow log level to be specified

sthapa:
This updates the charts to allow logging level for the web app to be specified


ivukotic (approved):
LGTM

servicex/templates/app/deployment.yaml
```diff
@@ -40,6 +40,10 @@ spec:
           value: "/opt/servicex/app.conf"
         - name: INSTANCE_NAME
           value: {{ .Release.Name }}
+    {{- if .Values.app.logLevel }}
```
BenGalewsky: Why would Log Level ever not be set? There is a default value in values.yaml... Is there a particular reason why anyone would want to leave this env var out?

BenGalewsky (changes requested):
Unless there is a reason to set the app.logLevel value to None I don't see a need to clutter up the chart with the test

servicex/templates/app/deployment.yaml
sthapa: No not really, but I prefer to have a check just in case there's some way that the things get here without the loglevel being set.  E.g. the admin has a logLevel:      line without a value.

BenGalewsky: By that logic we'd have to wrap every single environment variable with a check like this (which I don't suggest).

sthapa: I've removed the test

BenGalewsky (approved):
Thank you

PR #404, opened 2022-03-31
adding optional memcache container to rucio did finder pod

ivukotic:
this is needed to address issue: https://github.com/ssl-hep/ServiceX_DID_Finder_Rucio/issues/52

### DIFF
--- a/servicex/templates/did-finder-rucio/deployment.yaml
+++ b/servicex/templates/did-finder-rucio/deployment.yaml
@@ -23,7 +23,7 @@ spec:
         command: ["/usr/src/app/runme.sh"]
         imagePullPolicy: {{ .Values.didFinder.rucio.pullPolicy }}
         env:
-    {{- if .Values.secrets }}
+          {{- if .Values.secrets }}
           - name: RMQ_PASS
             valueFrom: 
               secretKeyRef:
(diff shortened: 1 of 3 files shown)


servicex/templates/did-finder-rucio/deployment.yaml
```diff
@@ -23,38 +23,46 @@ spec:
...
+          {{- end }}
         volumeMounts:
           - name: rucio-cfg
             mountPath: /opt/rucio/etc/
           - name: x509-secret
             mountPath: /etc/grid-security-ro
             readOnly: true
+      {{- if ((.Values.didFinder.rucio.memcache).enabled) }}
+      - name: {{ .Release.Name }}-did-finder-rucio-memcache
+        image: memcached:{{ .Values.didFinder.rucio.memcache.tag }}
```
BenGalewsky: Burt has requested that we don't have hardcoded docker images in our charts - he will want to use a local copy of this image so override it in `values.yaml`

BenGalewsky:
Do you really think memcache is needed here? It seems that there are simpler pure python implementations possible.

servicex/templates/did-finder-rucio/deployment.yaml
ivukotic: OK I understand.

BenGalewsky (changes requested):
Please update the reference documentation `docs/deployment/reference.md`

ivukotic:
> Please update the reference documentation `docs/deployment/reference.md`

done

BenGalewsky (approved):

PR #406, opened 2022-04-15
Remove preflight transformer from diagrams

sthapa:


BenGalewsky (approved):
Thanks for updating the docs!

PR #411, opened 2022-06-02
Template changes for minio cleanup

sthapa:
Template and servicex chart changes to support minio cleanup code


sthapa:
Somewhat of a draft PR for now, need to get the gh action to publish images automatically setup first

PR #420, opened 2022-07-06
Update chart to use latest version of bitnami postgres subchart

BenGalewsky:
# Problem
The bitnami Postgres sub chart was pinned to 8.3 - this chart has been deprecated and archived and no longer is available

# Approach
Update the dependency to the latest version of the sub chart. There are some changes to the names of the values we depend on. Updated `values.yaml` and the templates that pull these values. Also updated the example values files


sthapa (commented):
LGTM, I tried a quick test and this seems to work

sthapa (approved):

PR #421, opened 2022-07-27
Fix minio ingress reference

sthapa:
This fixes a reference to a value that's in the minio chart (the bitnami chart has it in a different location)


BenGalewsky (approved):
Great to get this cleaned up

PR #424, opened 2022-08-02
Fix reference to db

sthapa:
Fix template reference to postgresql db name


BenGalewsky (approved):
Thanks for catching this one

PR #425, opened 2022-08-19
Make the expected names of the user cert pem files clear

BenGalewsky:
# Problem
The docs don't say, but the names of the user cert pem  files is hardcoded in the script.

# Approach
Add to docs


ivukotic (approved):
this is important to know.

PR #434, opened 2022-08-29
Revert typo in memcached version

holzman:


sthapa (approved):
LGTM

BenGalewsky:
I wonder if the helm chart release script is bodging that? It tries to tag all of the docker images, but is only supposed to operate on our images, not external ones!

PR #454, opened 2022-09-23
Change defaults to move towards allowing default installs to work

sthapa:


BenGalewsky (approved):
Yep, makes sense

BenGalewsky:
@sthapa - please fix the conflicts

PR #459, opened 2022-10-04
Change default logging level and a lot of log lines.

ivukotic:
now logging is much more reasonable. Helped me discover new issues.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-    "python.linting.flake8Enabled": true,
-    "python.linting.enabled": true
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "python.linting.flake8Enabled": true,
+    "python.linting.enabled": true
+}
--- a/helm/servicex/values.yaml
+++ b/helm/servicex/values.yaml
@@ -15,8 +15,8 @@ app:
     tls:
       clusterIssuer: null
       enabled: false
-      secretName: '{{.Release.Name}}-app-tls'
-  logLevel: WARNING
+      secretName: "{{.Release.Name}}-app-tls"
+  logLevel: INFO
   mailgunApiKey: null
   mailgunDomain: null
   pullPolicy: Always
--- a/servicex_app/servicex/resources/internal/add_file_to_dataset.py
+++ b/servicex_app/servicex/resources/internal/add_file_to_dataset.py
@@ -44,8 +44,8 @@ class AddFileToDataset(ServiceXResource):
             # or a list of file dictionaries.
             add_file_request = request.get_json()
             submitted_request = TransformRequest.lookup(request_id)
-            current_app.logger.info(f"Submitted request: {submitted_request}",
-                                    extra={'requestId': request_id})
+            current_app.logger.debug(f"Adding files to request: {submitted_request}",
+                                     extra={'requestId': request_id})
 
             # check if the request is bulk or single file
             if type(add_file_request) is dict:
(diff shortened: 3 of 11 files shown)


helm/servicex/values.yaml
```diff
@@ -15,8 +15,8 @@ app:
     tls:
       clusterIssuer: null
       enabled: false
-      secretName: '{{.Release.Name}}-app-tls'
-  logLevel: WARNING
+      secretName: "{{.Release.Name}}-app-tls"
+  logLevel: INFO
```
sthapa, line 19: We shouldn't be changing log levels in the code.  You can set the levels using the LOG_LEVEL env var for the servicex container or by setting flask to debug mode.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
sthapa: We shouldn't be checking in vscode or other IDE files

servicex_app/servicex/resources/internal/add_file_to_dataset.py
```diff
@@ -61,8 +61,11 @@ def put(self, request_id):
                 self.lookup_result_processor.add_file_to_dataset(submitted_request, db_record)
 
             db.session.commit()
-            current_app.logger.info(f"Got {len(add_file_request)} files.",
-                                    extra={'requestId': request_id})
+            current_app.logger.info("Adding files.",
+                                    extra={
+                                        'requestId': request_id,
+                                        'nfiles': len(add_file_request)
```
sthapa, line 67: I don't think the nfiles will get included in the log output

ivukotic: it won't. But I don't want any of the extra things included. It does get sent to logstash where it matters.

.vscode/settings.json
ivukotic: agree. removed.

helm/servicex/values.yaml
ivukotic: but this is not in the code. This is simply a change of the default in values.yaml?

sthapa: Or in the default deployment.  I believe the convention is to use WARNING/WARN as the default to avoid cluttering up the logs with extraneous messages.  We should stick to the convention and then increase it in our testing instances as needed.

servicex_app/servicex/resources/internal/add_file_to_dataset.py
sthapa: Okay, but what about admins not using logstash?  We need to also consider what admins that are only going to get the console output need in order to debug problems.  I'd rather not have to start spinning up a special version of the container with logging changes in order to debug an issue at BNL or UNL or with Burt's instance.

helm/servicex/values.yaml
ivukotic: Until we get the code debugged, and performance understood, I would really prefer if we kept default on INFO, and logstash monitoring on by default. We won't have too many messages as there are not many users. Once that's done, I completely agree that WARN would be better.

sthapa: Well, the default values can be overridden in the environment of the container.  If we need more information, why not update the values.yaml in our flux deployments to set the log level to DEBUG?  Also, it sounds like we should also be enabling flask debugging as well in our deployments.  This is something that's relatively easy to do without having to change the default.yaml for the chart.

servicex_app/servicex/resources/internal/add_file_to_dataset.py
ivukotic: I have added extra fields in output. 
but I would really prefer if people would just leave the default settings and use ES to debug things. It is much easier when you can for example select a requestID or source file and get only these lines.

sthapa (approved):
LGTM

Issue #417, opened 2022-06-13
Create Code Generator Library


BenGalewsky:
_As a code generator developer I want to have a library of common functionality so I don't have to repeat myself_

The code generators share a fair amount of common code. More shared features are desired. Create a new library and refactor all of the existing code generators to use this library

PR #464, opened 2022-10-06
Migrate Code Generators to CodeGen Lib

BenGalewsky:
# Problem
There is duplicated code across the various CodeGenerators

Solves #417 

# Approach
Developed a [library](https://github.com/ssl-hep/ServiceX_Code_Generator_lib) that contains this code. Interfaced this code generator to the lib and deleted the overlap.


gordonwatts (approved):
This looks great! And so nice to see so many deletes!

PR #465, opened 2022-10-10
Uproot transform return fix

ivukotic:
Now Uproot transform returns number of events and size of the output. This also prepares xAOD transform for new logging and ups version of AnalysisBase.

### DIFF
--- a/xaod_cpp_transformer/requirements.txt
+++ b/xaod_cpp_transformer/requirements.txt
@@ -10,9 +10,7 @@ pyarrow==0.16.0
 pytest==4.6.11
 pytest-mock==2.0.0
 requests==2.24.0
-# servicex-transformer==1.0.3
-# uproot==3.11.7
-servicex-transformer==1.0.4-rc1
+servicex-transformer==1.0.4
 xxhash==1.4.4
 psutil==5.8.0
 python-logstash == 0.4.8
--- a/xaod_cpp_transformer/requirements.txt
+++ b/xaod_cpp_transformer/requirements.txt
@@ -10,7 +10,9 @@ pyarrow==0.16.0
 pytest==4.6.11
 pytest-mock==2.0.0
 requests==2.24.0
-servicex-transformer==1.0.3
-uproot==3.11.7
+# servicex-transformer==1.0.3
+# uproot==3.11.7
+servicex-transformer==1.0.4-rc1
 xxhash==1.4.4
 psutil==5.8.0
+python-logstash == 0.4.8
--- a/xaod_cpp_transformer/transformer.py
+++ b/xaod_cpp_transformer/transformer.py
@@ -27,6 +27,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import json
 import logging
+import logstash
 import os
 import time
 import timeit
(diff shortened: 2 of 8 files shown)


xaod_cpp_transformer/transformer.py
```diff
@@ -105,7 +181,7 @@ class TimeTuple(namedtuple("TimeTupleInit", ["user", "system", "iowait"])):
     # system: float
     # iowait: float
 
-    @property
+    @ property
```
sthapa: Is this a typo? It's valid syntax but different from the usage in the existing code

ivukotic: yes.fixed.

sthapa:
Looks fairly reasonable, I had a small comment about the decorator formatting and unless there's a good reason, probably cleaner to delete code rather than commenting out (it happens in a few places but is pretty minor overall).

xaod_cpp_transformer/requirements.txt
```diff
@@ -10,7 +10,9 @@ pyarrow==0.16.0
 pytest==4.6.11
 pytest-mock==2.0.0
 requests==2.24.0
-servicex-transformer==1.0.3
-uproot==3.11.7
+# servicex-transformer==1.0.3
```
BenGalewsky: Any reason this can't just be 1.0.4? Why are we depending on an RC of this library?

xaod_cpp_transformer/requirements.txt
```diff
@@ -10,7 +10,9 @@ pyarrow==0.16.0
...
+# uproot==3.11.7
```
BenGalewsky: Please don't comment out code. We don't need uproot in this transformer, so just delete the requirement

xaod_cpp_transformer/transformer.py
```diff
@@ -27,22 +27,22 @@
...
+# import uproot
 
 from servicex.transformer.servicex_adapter import ServiceXAdapter
 from servicex.transformer.transformer_argument_parser import TransformerArgumentParser
 from servicex.transformer.object_store_manager import ObjectStoreManager
 from servicex.transformer.rabbit_mq_manager import RabbitMQManager
-from servicex.transformer.uproot_events import UprootEvents
-from servicex.transformer.uproot_transformer import UprootTransformer
-from servicex.transformer.arrow_writer import ArrowWriter
+# from servicex.transformer.uproot_events import UprootEvents
```
BenGalewsky: Please don't comment out code. We have the git commit history if for some reason we need this back

xaod_cpp_transformer/transformer.py
```diff
@@ -51,8 +51,64 @@
...
+
+#         string = super().format(record)
+#         extra = {k: v for k, v in record.__dict__.items()
+#                  if k not in self.def_keys}
+#         if len(extra) > 0:
+#             string += " extra: " + str(extra)
+#         return string
+
+
+class LogstashFormatter(logstash.formatter.LogstashFormatterBase):
```
BenGalewsky, line 78: Can this not be in the transformer library? Is it specific to the xaod transformer?

xaod_cpp_transformer/transformer.py
```diff
@@ -91,8 +163,8 @@ def parse_output_logs(logfile):
         matches = events_processed_re.finditer(buf)
         for m in matches:
             events_processed = int(m.group(1))
-        logger.info("{} events processed out of {} total events".format(
-            events_processed, total_events))
+        # logger.info("{} events processed out of {} total events".format(
```
BenGalewsky: more commented out code. Just delete this if we don't need it

BenGalewsky (changes requested):
Please don't comment out code. It makes the code base noisy and we have git commit history if we need it back.
The LogStashFormatter sure looks like it could be in the library so we don't have to repeat it in the uproot transformer.
Please release a fill 1.0.4 transformer library if it is correct. We shouldn't be releaseing the transformer with a dependency on a RC lib unless there is a very good reason to

xaod_cpp_transformer/requirements.txt
ivukotic: agree. Will cut 1.0.4

xaod_cpp_transformer/requirements.txt
ivukotic: agree

xaod_cpp_transformer/transformer.py
ivukotic: agree

xaod_cpp_transformer/transformer.py
ivukotic: It could. There is nothing specific to xaod transformer.  But in transformer library:
```
    def __init__(self, servicex_endpoint, logger=None):
        # Default logger doesn't print so that code that uses library
        # can override
```
Later we can think how to do this centrally for all the services.

Issue #466, opened 2022-10-10
Posgresql deployment needs fixing


ivukotic:
Postgresql often uses > 1500 mcpus. 

In its pod description I find this:

 ```      
        app.kubernetes.io/name=postgresql
        helm.sh/chart=postgresql-11.6.26
  Service Account:  default
  Containers:
   postgresql:
    Image:      docker.io/bitnami/postgresql:14.4.0-debian-11-r23
    Port:       5432/TCP
    Host Port:  0/TCP
    Requests:
      cpu:      250m
      memory:   256Mi
```

There are two problems here: 
a) it needs to request 2 CPUs otherwise it will be kicked out once we have a lot of load from Transformers.
b) versions need to be fixed. We know we are deploying 14.4 but helm claims it is 11.6.26.

sthapa:
For b, the helm chart version is 11.6.26.  However the underlying app version is 14.4.   There's a chart parameter that lets us change the version of postgresql that will be installed.  However, given that we've already released using 14.4, it'd be pretty disruptive to go back to 11.6.

sthapa:
For a, that's the default setting from the postgresql subchart, a quick fix for this would be to update the defaults in the servicex chart.

PR #467, opened 2022-10-10
update postgresql default cpu request

ivukotic:
addressing https://github.com/ssl-hep/ServiceX/issues/466

### DIFF
--- a/helm/servicex/values.yaml
+++ b/helm/servicex/values.yaml
@@ -88,6 +88,9 @@ postgresql:
         database: servicex
         postgresPassword: leftfoot1
   primary:
+    resources:
+      requests:
+        cpu: 2
     persistence:
       enabled: false
 rabbitmq:
--- a/helm/servicex/values.yaml
+++ b/helm/servicex/values.yaml
@@ -90,7 +90,8 @@ postgresql:
   primary:
     resources:
       requests:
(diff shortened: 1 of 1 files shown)


sthapa:
This should do the trick based on the bitnami docs but would need testing

helm/servicex/values.yaml
```diff
@@ -88,6 +88,9 @@ postgresql:
         database: servicex
         postgresPassword: leftfoot1
   primary:
+    resources:
+      requests:
+        cpu: 2
```
BenGalewsky: I really like keeping the default `values.yaml` useful for developers or CI jobs. I think. this belongs in the production flux configs and not here.

ivukotic: is cpu:1 OK with you? Or at least 500m ?

BenGalewsky: I don't have any problem with the bitnami default
```
  resources:
    limits: {}
    requests:
      memory: 256Mi
      cpu: 250m
```

That's good for developers and anyone else should really do what makes sense for their production deployment

ivukotic: do you mind if we still spell 250m here, it is not very obvious where this is set, and even that this is an issue ?

PR #468, opened 2022-10-12
Web frontend changes

ivukotic:
now it shows correct number of files to process,
correctly calculate percentage completed.
dashboard is now in the main menu.
added spinner
reorganized queries so it does not ask for deployments that don't exist.


servicex_app/servicex/templates/base.html
```diff
@@ -48,30 +48,39 @@
             {% endif %}
           </ul>
 
-          {% if config["ENABLE_AUTH"] %}
```
BenGalewsky, line 51: Can you confirm that you have tested this both with Auth enabled and without? I could imagine there are some funny cases that could be triggered

ivukotic: just tested it on River int xaod instance and ti looks fine.

PR #471, opened 2022-10-17
Adding monitoring

ivukotic:
it adds monitoring link,
updates version of setup-action
fixes xaod transformer


sthapa:
This looks reasonable.   However the github action changes should probably be in another PR since that seems unrelated to adding a monitoring dash to servicex.

PR #475, opened 2022-10-23
Allow uproot transformer to produce ROOT files directly instead of parquet

ponyisi:
Allow the uproot transformer to
* write results directly to ROOT files instead of parquet
* set the tree name in the ROOT file from the tree-name property (can be useful if we're processing systematics trees, then the client can merge the resulting files without confusion)

I think I'm overloading the "tree-name" request parameter, but I get the impression that its meaning is a bit ill-defined now. I don't think there's any coherent mechanism for passing tree name information to the different components right now (e.g. the frontend libraries don't pass the info, the top-level transformer doesn't have access to the knowledge from the qastle because it's hidden in the python from the code generator). It's probably worth having a discussion what we want to do there because once we consider ROOT and not Parquet it becomes a meaningful degree of freedom.

### DIFF
--- a/.github/workflows/ci_servicex.yaml
+++ b/.github/workflows/ci_servicex.yaml
@@ -89,7 +89,7 @@ jobs:
         uses: elgohr/Publish-Docker-Github-Action@master
         with:
           context: ${{ matrix.app.dir_name }}
-          name: sslhep/${{ matrix.app.image_name }}:${{ steps.extract_tag_name.outputs.imagetag }}
+          name: ponyisi/${{ matrix.app.image_name }}:${{ steps.extract_tag_name.outputs.imagetag }}
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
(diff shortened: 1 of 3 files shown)


ponyisi:
To test

```
from servicex import ServiceXDataset, ignore_cache
from func_adl_servicex import ServiceXSourceUpROOT

dataset_uproot = "user.kchoi:user.kchoi.ttHML_80fb_ttbar"
uproot_transformer_image = "ponyisi/servicex_func_adl_uproot_transformer:uproot-to-root"

sx_dataset = ServiceXDataset(dataset_uproot, image=uproot_transformer_image, backend_name="uproot")
ds = ServiceXSourceUpROOT(sx_dataset, "nominal")
ds.return_qastle = True
data = (ds.Select("lambda e: { 'lep_pt_1': e.lep_Pt_1,'lep_pt_2': e.lep_Pt_2 }")).value()

with ignore_cache():
    files = sx_dataset.get_data_rootfiles(data)
    print(files)
```

ponyisi:
Superseded

Issue #302, opened 2021-06-10
Don't require ServiceX Code in the transformer container


gordonwatts:
# Story
_As a transform author, I want to write a new transformer that employs an existing experiment's container without having to inject the transformer library into that container so that I can use very old containers to do analysis on data that has been preserved from previous releases of software._

# Assumptions
1. The app creates pods with two containers: The ServiceX driver and the Physics image.
2. The two containers communicate via a shared `emptyDir` volume
3. The ServiceX driver container will write a file-based FIFO queue
4. The Physics container will write the output file to the shared volume along with a semaphore file
5. The ServiceX driver container will wait for Semaphore files and transfer the output file to Minio and signal ServiceX
6. There should not be any python code in the Physics container. The workflow should use Bash and simply invoke the transformer command
7. The Code Generator should be modified to return the command line to be invoked in the physics container.

# Context
This is the result of noticing two things recently when dealing with our transformers:

* When @sthapa was upgrading logging on the ATLAS xAOD Release 20 containers, he was constantly fighting with Python 2 vs Python 3. Eventually, with a good deal of time on his part, he was able to make the logging sections of the logging library work in both cases.
* When @gordonwatts built the CMS Run 1 AOD transformer, it turned out to be against a container image of RHEL6 - which is no longer supported by CERN. It is so unsupported by CERN, in fact, that yum's index of OS package repositories no longer exists on the web, meaning yum is totally broken. Further, the image does not have `pip` installed on it: so it was not possible to install `pip` and thus to properly install the python packages to run the transformer. In the end, I downloaded python 3.8, and thank goodness is builds under gcc 4.X (which IS VERY OLD), and I was able to bootstrap up from there. But since Python 2 and Python 3 now have to co-exist, and I had to do some hacky, but reproducible, things. The solution works, but it is quite fragile.

Both of these things lead me to wonder if there wasn't a better way.

gordonwatts:
A possible approach - which I don't want to add above because I don't know how crazy it is - would be to have a transform manager container that runs configures containers/jobs to run, and each one just processes a single file off of rabbit, but it has to mount the output file, and when the output file is written, it has to be sent to the proper place (minio, etc.).

Or perhaps a container that runs other via docker?

I don't really know what is possible here.

[...]

PR #476, opened 2022-10-24
Don't require ServiceX Code in Science Containers

BenGalewsky:
# Problem
ServiceX transformers require python code that runs inside the transformer container. Experiment-provided images may not have python 3 installed and there can be complicated dependencies to work through.

Solution to #302 

# Approach
Introduce a sidecar container to run alongside the science image. This container holds all of the python and ServiceX dependencies and communicates with the science container solely through a shared volume. The sidecar container copies a bash script into this shared volume and the science container simply executes this script.

# Basic Workflow
1. The code generators now include a top-level script that will attempt to transform a single file. It is a python script in the uproot transformer, and a bash script for the xAOD transformer. The script it copied into the generated code. The code generators also now have a [transformer_capabilities.json](https://github.com/ssl-hep/ServiceX/blob/sidecar-monorepo/code_generator_funcadl_uproot/transformer_capabilities.json) to communicate features to the sidecar about how to interact with the transform_single_file script.
2. The App's TransformerManager creates two containers for each transformer pod. The [sidecar container](https://github.com/ssl-hep/ServiceX/blob/0196064d4fe59edc954b6db4d0522fe529312d52/servicex_app/servicex/transformer_manager.py#L194) and the [transformer container](https://github.com/ssl-hep/ServiceX/blob/0196064d4fe59edc954b6db4d0522fe529312d52/servicex_app/servicex/transformer_manager.py#L181).
3. It creates an EmptyDir volume and mounts it into both containers
4. The Sidecar container launches and runs its [transformer.py](https://github.com/ssl-hep/ServiceX/blob/sidecar-monorepo/transformer_sidecar/src/transformer_sidecar/transformer.py) script.
5. This script copies [watch.sh](https://github.com/ssl-hep/ServiceX/blob/sidecar-monorepo/transformer_sidecar/scripts/watch.sh) to the shared folder.
6. The science container waits for watch.sh to appear and then begins executing it.
7. The transformer script creates a directory in the shared volume named after the transformation request ID. The watch script will wait for commands in that directory.
8. When a RabbitMQ message is received, the transformer script saves the record to the shared folder as a JSON file.
9. The watch script in the science container waits for these json files. It uses the values in this file to launch the local script to transform a single file.
10. The watch script checks the return value from the transform single file script and writes a signal file to the shared folder. It ends with `.done` to signal a successful transformation, or `.failed` for failure.
11. All of the output from the file transform are written to a `.log` file in the shared volume
12. The transformer script monitors the shared volume for new output files and these signal files.
13. Any new output files are uploaded to object store 
14. The log files are scanned for metrics of the request, or clues about failures.

# Summary of Changes
## Helm Chart
The helm chart has been updated to change the way the transformer is configured. 
New values to specify the sidecar container:
- sidecarImage
- sidecarTag
- sidecarPullPolicy

Values to specify the Science container:
- defaultScienceContainerImage
- defaultScienceContainerTag
- scienceContainerPullPolicy

These can still be overridden in the transform request to specify some specific version of the transformer.

The watch script takes three arguments: A language, a script, and the path to monitor. The language can be `python` or `bash`, and the script is the path to the generated transform_single_file script.
- transformer.language
- transformer.exec
- transformer.outputDir

## ServiceX App
The config file for the app has been updated to manage the new values from the helm chart.

The TransformerManager has been updated to create the shared volume and the two containers.

## Code Generators
The code generators are now responsible for writing a `transform_single_file` script into the generated code. This driver code was previously in the individual transformer containers. It can be a python script or a bash script.

The code generator also places a json file in the generated code that records the transformer's basic capabilities. Right now this is used to specify the class of log file parser that can make sense of the output. In the future we will let the sidecar know what types of files the transformer can natively generate.

## The Sidecar Container
This is the big addition to the service. It has a main script that is executed when the container starts up. When a transform request is received, the script creates two threads. One [watches the output directory](https://github.com/ssl-hep/ServiceX/blob/sidecar-monorepo/transformer_sidecar/src/transformer_sidecar/watched_directory.py) and the other [uploads results to the object store](https://github.com/ssl-hep/ServiceX/blob/sidecar-monorepo/transformer_sidecar/src/transformer_sidecar/object_store_uploader.py).

The transformer library is deprecated with this release and the classes have been moved directly into this repo.

The `WatchedDirectory` thread waits for a `.done` or `.failed` signal file, sets internal state and tells to object store uploader to shut down once the last queued file is uploaded.

The `transformer_capabilities.json` file has an entry for `stats-parser` which specifies the subclass of [TransformerStats](https://github.com/ssl-hep/ServiceX/blob/0196064d4fe59edc954b6db4d0522fe529312d52/transformer_sidecar/src/transformer_sidecar/transformer_stats/__init__.py#L33) to instantiate to parse the resulting log file. It can find the total number of events and file size. If there is an exception, it will attempt to narrow down exactly what the cause was.


code_generator_funcadl_uproot/servicex/templates/transform_single_file.py
```diff
@@ -0,0 +1,59 @@
+import os
```
gordonwatts, line 1: What roll is the template directory playing here?

helm/servicex/templates/app/configmap.yaml
```diff
@@ -96,8 +94,17 @@ data:
     {{- end }}
     TRANSFORMER_VALIDATE_DOCKER_IMAGE = {{- ternary "True" "False" .Values.app.validateTransformerImage }}
     TRANSFORMER_MESSAGING = 'none'
-    TRANSFORMER_DEFAULT_IMAGE = "{{ .Values.transformer.defaultTransformerImage }}:{{ .Values.transformer.defaultTransformerTag }}"
     
+    TRANSFORMER_SIDECAR_IMAGE = "{{ .Values.transformer.sidecarImage }}:{{ .Values.transformer.sidecarTag }}"
```
gordonwatts, line 98: When we get to the point we want to set the side car in the code generator, how will that work? Or are we just going to ignore that for this PR?

transformer_sidecar/src/transformer_sidecar/transformer_logging/logstash_formatter.py
```diff
@@ -0,0 +1,61 @@
+# Copyright (c) 2022, IRIS-HEP
```
gordonwatts, line 1: Is this cut/paste code from somewhere? If so, should we have a common location for it (in some library)?

transformer_sidecar/src/transformer_sidecar/transformer_stats/aod_stats.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) 2022, IRIS-HEP
```
gordonwatts, line 1: What is the development path when I come along with a new transformer and need it to scan the output? Does that mean modifying this directory to add something? Should we either generalize things with regex's or put something in the codegen libraries that it then ships as part of the code bundle that each transformer runs?

transformer_sidecar/src/transformer_sidecar/watched_directory.py
```diff
@@ -0,0 +1,139 @@
...
+            self.watched_directory.stop(success=False)
+
+            # Terminate the upload thread too
+            self.result_upload_queue.put(ObjectStoreUploader.WorkQueueItem(None))
+            return
+
+        self.logger.info('File {fn} created.'.format(fn=event.src_path))
+
+        # check if file still being written/copied
+        while True:
```
gordonwatts, line 117: Is this going to work when I/O slows to a crawl? We've seen this with some GRID sites - they plow along and then suddenly hang for minutes without any I/O and then pick up again. What will that do here?

transformer_sidecar/src/transformer_sidecar/watched_directory.py
```diff
@@ -0,0 +1,139 @@
...
+        # Nothing to do on directory events or log files
+        if event.is_directory or event.src_path.endswith('.log'):
+            return
+
+        # Look for signal that the job is done
+        if event.src_path.endswith(".done"):
+            self.watched_directory.stop(success=True)
+
+            # Terminate the upload thread too
+            self.result_upload_queue.put(ObjectStoreUploader.WorkQueueItem(None))
```
gordonwatts, line 103: Below you protect this with a try/except (at bottom of this method)... but not here. Does one need some sort of protection here?

code_generator_funcadl_xAOD/servicex/__init__.py
```diff
@@ -1,27 +0,0 @@
-# Copyright (c) 2019, IRIS-HEP
```
gordonwatts: You aren't deleting this file but removed the copyright notice?

gordonwatts (commented):
* Under `./transformer_sidecar/tests` the `__pycache__` directory is checked in.

code_generator_funcadl_uproot/servicex/templates/transform_single_file.py
BenGalewsky: This is where we can store files that will be copied directly into the tarball. All the rest of the code is emitted from `func_adl_xAOD`. It makes sense that this file is part of ServiceX since it is intimately tied to our code and is of no use outside of ServiceX

helm/servicex/templates/app/configmap.yaml
BenGalewsky: This PR is already big enough! This is basically keeping what we currently have, but slightly renaming things

transformer_sidecar/src/transformer_sidecar/transformer_logging/logstash_formatter.py
BenGalewsky: It is no longer cut and paste as far as I know. This is unique to the transformers. @ivukotic can tell us for sure

transformer_sidecar/src/transformer_sidecar/transformer_stats/aod_stats.py
BenGalewsky: For now it will require a new class here and a reference to it in the transformer's `transformer_capabilities.json`
The next time we need to do this, we can burden that issue with generalizing this. Again, this PR is already way too ambitious

transformer_sidecar/src/transformer_sidecar/watched_directory.py
BenGalewsky: We might need to write to a scratch file and then do a `mv` to this directory

transformer_sidecar/src/transformer_sidecar/watched_directory.py
BenGalewsky: Actually the try/catch is useless since in the catch it will try to put something to the queue (which is what caused the exception in the first place) - I deleted that try/catch and just let the exception kill the watched directory

code_generator_funcadl_xAOD/servicex/__init__.py
BenGalewsky: Restored the copyright notice. We do need the __init__.py

transformer_sidecar/src/transformer_sidecar/transformer_stats/aod_stats.py
gordonwatts: This sounds ok for now - but perhaps a note could be added somewhere in the docs for future people so they don't get too lost trying to figure this out?

transformer_sidecar/src/transformer_sidecar/watched_directory.py
gordonwatts: If this is required, we'll need to change how the scripts in all the transformers work. In the case of the C++. We should start a new bug/feature in those so we can get to work if this is really going to be needed.

BenGalewsky:
Congratulations, @Michael-D-Johnson your work has been merged!

PR #482, opened 2022-10-27
Migrate to Poetry and Flask 2

### LINKED ISSUE #390: Explicitly Manage Python Dependencies
We've had two instances in the last 2 weeks or so where a release broke because an underlying python module was updated and that caused problems.  We need to pin dependencies or have a mechanism for doing so without letting modules get too stale.

### LINKED ISSUE #456: Upgrade to Flask2
# Problem
Flask1 has been deprecated and we have to go to great lengths to find versions of libraries that still work together with it.

# Acceptance Criteria
1. Servicex_app, code_generator_funcadl_uproot, code_generator_funcadl_xAOD, code_generator_python all use flask 2
2. Any tests succeed

# Assumptions
1. Review to see  if Flask-RESTful is compatible with flask2 and propose alternate solution if not
2. Clean up the pinned dependencies

BenGalewsky:
# Problem(s)
Our builds are not reproducible since packages listed in `requirements.txt` are not pinned. 

The system is very fragile since we are using an EOL version of Python and and EOL version of Flask.

Ideally we would want to split these into two PRs, but the old version of python doesn't support the current version of Poetry,  and the modern dependency resolvers can't thread the needle to get working versions of the libraries.

Solves #390 and #456 

# Approach
1. Go through each service and replace `requirements.txt` with a `pyproject.toml` and generate a `poetry.lock`
2. Update the CI Workflow to install poetry and load dependencies with it
3. Update the docker files to install poetry and load in dependencies with it.
4. Update the ServiceX app and the code generators to use Flask2
5. Fix a few errors in the tests that came up with the new version of flask.
6. Discovered that a few services really can't be upgraded and added issues to the backlog to fix them separately.


.github/workflows/ci_servicex.yaml
```diff
@@ -51,25 +51,24 @@ jobs:
         if: ${{ matrix.app.test_required }}
         run: |
           echo "${{ matrix.app }}"
-          python -m pip install --upgrade pip setuptools wheel
-          pip install --no-cache-dir -r requirements.txt
-          pip install --no-cache-dir -r requirements_test.txt
+          python -m pip install --upgrade pip poetry
```
BenGalewsky: We need to pin poetry itself! There is a GitHub action for installing poetry:
```
      - name: Install poetry
        uses: abatilo/actions-poetry@v2
        with:
          poetry-version: 1.2.2
```

code_generator_funcadl_uproot/pyproject.toml
```diff
@@ -0,0 +1,21 @@
...
+name = "code-generator-funcadl-uproot"
+version = "0.1.0"
+description = ""
+authors = ["Your Name <you@example.com>"]
+readme = "README.md"
+packages = [{include = "code_generator_funcadl_uproot"}]
+
+[tool.poetry.dependencies]
+python = "~3.7"
+servicex-code-gen-lib = "1.1.0a1"
```
BenGalewsky: Let's cut a real 1.1.0 release of this library

servicex_app/pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+passlib = "^1.7.4"
+flask-sqlalchemy = "^3.0.1"
+kubernetes = "^24.2.0"
+globus-sdk = "^3.12.0"
+cryptography = "^38.0.2"
+Bootstrap-Flask = "^2.1.0"
+humanize = "^4.4.0"
+python-logstash = "^0.4.8"
+gunicorn = "^20.1.0"
+pytest = "^7.1.3"
```
BenGalewsky: These should be in a test group and not bundled in with the production deploy

servicex_app/tests/resources/transformation/test_get_all.py
```diff
@@ -36,6 +36,14 @@ class TestAllTransformationRequest(ResourceTestBase):
     def example_json():
         return [{'request_id': '123'}, {'request_id': '456'}]
 
+    @staticmethod
```
BenGalewsky: I think this can be a fixture or a method in `ResourceTestBase`
There are a lot of these copy/pasted `fake_header` methods. They should all be cleaned up

servicex_app/tests/test_decorators.py
```diff
@@ -50,7 +54,7 @@ def test_oauth_decorator_saved(self, client, user, captured_templates):
         assert template.name == "profile.html"
         assert context["user"] == user
 
-    def test_auth_decorator_auth_disabled(self, client):
+    '''def test_auth_decorator_auth_disabled(self, client):
```
BenGalewsky: Don't comment out code. If we don't need this then delete it from the commit

shriram192:
After moving to python 3.10 with poetry 1.2.2, these are the issues present with the repos

1. Rucio Image does not build with poetry - affected repos are 
    1. did_finder_rucio
    2. x509_secrets
2. Pypi Package servicex_transformer does not build with 3.10 - affected repos are
    1. transformer_sidecar
    2. uproot_transformer
    3. xaod_cpp_transformer

.github/workflows/ci_servicex.yaml
shriram192: The CI has been updated and its working!

code_generator_funcadl_uproot/pyproject.toml
shriram192: Screenshot 2022-10-31 at 1 52 24 PM

.github/workflows/ci_servicex.yaml
shriram192: Screenshot 2022-10-31 at 1 52 24 PM

code_generator_funcadl_uproot/pyproject.toml
shriram192: Screenshot 2022-10-31 at 1 58 14 PM

The new release can be found on pypi

servicex_app/pyproject.toml
shriram192: test groups have been created for each repository

servicex_app/tests/test_decorators.py
shriram192: Understood. I have removed those test cases.

servicex_app/tests/resources/transformation/test_get_all.py
shriram192: The methods have been moved to ResourceTestBase and WebTestBase

PR #485, opened 2022-10-29
Custom base image that has rucio client and voms things working with python3.9

ivukotic:
it is not 3.10 but I tested it and it works.

### DIFF
--- a/.github/workflows/ci_servicex.yaml
+++ b/.github/workflows/ci_servicex.yaml
@@ -87,7 +87,7 @@ jobs:
           dockerfile: ${{ matrix.app.dockerfile }}
 
   publish-servicex:
-    needs: [build-matrix, test-servicex]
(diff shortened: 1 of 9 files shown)


BenGalewsky (commented):
This is amazing, @ivukotic 
The X509 service needs most of this. What do you think about moving the base to its own repo and then we build the rucio_did finder and the X509 service off that image?

ivukotic:
That’s what I already did except I kept base image a part of monorepo. The only thing I didn’t do is tag it in the same way as other images. I left a build argument in the DID and x509 dockerfiles if you want to do it.

I

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

BenGalewsky:
Closed in favor of #487 and [new repo with docker image](https://github.com/ssl-hep/rucio-client)

PR #486, opened 2022-10-30
Fix up the uproot getting started

gordonwatts:
Update the `uproot` query in getting started:

* Use OpenData
* Use a more sophisticated query to help ease user into best practices

### DIFF
--- a/docs/user/getting-started.md
+++ b/docs/user/getting-started.md
@@ -100,38 +100,42 @@ entry
 [11355980 rows x 1 columns]
 ```
 
-### Uproot
+### uproot
+
+Instead of a rucio dataset, here we will use a file directly available over `https`,
+and a slightly more complex query, and we'll ask for the data to be locally downloaded
+so we can access the files directly.
 
 ```python
-import pandas as pd
 from func_adl_servicex import ServiceXSourceUpROOT
+import servicex
+import awkward as ak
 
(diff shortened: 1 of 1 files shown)


docs/user/getting-started.md
```diff
@@ -100,38 +100,42 @@ entry
...
-data = src.Select("lambda e: {'JetPT': e['AnalysisJetsAuxDyn.pt']}") \
-    .AsParquetFiles('junk.parquet') \
-    .value()
-df = pd.read_parquet(data[0])
-print(df)
+import servicex
+import awkward as ak
+
+dataset_name = [
+    "https://xrootd-local.unl.edu:1094//store/user/AGC/nanoAOD/nanoaod15.root"
```
alexander-held, line 115: While I am not aware of any intention of moving this file, I am wondering whether in the longer term it would be safer to use [scikit-hep-testdata](https://github.com/scikit-hep/scikit-hep-testdata/) (or an IRIS-HEP version of such a repository). A potential issue of the file disappearing would be alleviated by CI of course that could catch it when it happens.

BenGalewsky (approved):
Seems more robust too

docs/user/getting-started.md
matthewfeickert: Strong agree here.

PR #487, opened 2022-11-01
Migrate to sslhep/rucio-client to get modern version of python

BenGalewsky:
# Problem
The official rucio-client image is based on a very old version of python.

# Approach
* Switch to a [locally built docker image](https://github.com/ssl-hep/rucio-client)
* Upgrade to poetry to match the goals of 1.0.34
* Clean up install of libraries that have been picked up in the base image already


ivukotic (approved):
LGTM

PR #488, opened 2022-11-01
RUCIO DID Finder now listens to ?Files

gordonwatts:
New DID Finder Lib Version:

* Will correctly handle `?Files` when a bulk request comes through
* Will send batches of files all at once when it has an oportunity.


BenGalewsky (approved):
It would have been a nice touch to link this with the PR in the DID Finder lib since there is nothing really to review here.
So glad this has been fixed!

PR #490, opened 2022-11-02
Helm ci

prajwalkkumar:
This PR has the required changes that 
1. Do not erase the history of index.yaml during creation of a new release
2. Use yq instead of pyyaml, as it has native support for inplace replacement of yaml file values and has a nicer way to preserve comments in the yaml file. yq also preserves strings with "" in the yaml file values.


.github/workflows/values.yaml
```diff
@@ -0,0 +1,190 @@
+---
```
BenGalewsky: Is this file supposed to be in the PR?

prajwalkkumar:
Deleted values.yaml

BenGalewsky (approved):
Neat! Nice find

PR #491, opened 2022-11-02
Simplify names of Rucio DID Finder containers

BenGalewsky:
# Problem
Even though pod container names are namespaces by the pod, the Rucio DID Finder's two containers have names that include the helm deployment name. This is cumbersome when you want to read the logs for the DID finder since you have to specify which container with a `-c` argument.

# Approach
Removed the helm deployment from the container names


ivukotic (commented):
is "did-finder" sufficient? If one instance has both did-finder-rucio and did-finder-opendata?

BenGalewsky:
> is "did-finder" sufficient? If one instance has both did-finder-rucio and did-finder-opendata?

These are containers inside the `did-finder-rucio` pod, so there is no chance for confusion

PR #493, opened 2022-11-04
Reformat GHA to put more info before the ... in the job status page

BenGalewsky:
# Problem
When the system builds, the GitHub action page shows the various steps, but They mostly show _Servicex-test, ServiceX_c..._ - it's not terribly useful. 

# Approach
Re-ordered entries in the JSON file so the directory name is first.

Also shortened the job names so they don't have the redundant _ServiceX_ taking up precious space


prajwalkkumar (approved):
LGTM

Issue #494, opened 2022-11-07
remove table file_status


ivukotic:
This table is not used anymore. 
We should remove the table and all the code that references it.

Issue #460, opened 2022-10-04
There is a race condition in TransformerFileComplete. It tries to "transform_complete" a request multiple times.


ivukotic:

ivukotic:
transformer_file_complete is called any time a file is completed. it adds to the DB a record of finished file. then it checks if there are any remaining files to be done and if not tries to finalize the request and clean up related k8s resources. But since flask is multithreaded, the function gets called from two or more threads. This causes exceptions when trying to clean up k8s resources that don't exist any more.

ivukotic:
solved in https://github.com/ssl-hep/ServiceX/pull/495

PR #495, opened 2022-11-09
Db cleanup

ivukotic:
This turned into much more then removing unused table and solves issues: #494, #460. It also reduces load on Postgresql factor 20.
Main changes:

- drops file_status table
- stops filling files table as it is not used at all and slows down things
- to requests table adds columns files_complete, files_failed, and drops files_skipped 
- adds indices to transform_request table
- solves race conditions in accounting for completed files
- now refreshing of dashboard is much faster and less database intensive
- commented out a bunch of tests as they will have to be anyhow rewritten in the next phase of DB reorganization.


BenGalewsky:
There is a lot of interesting stuff here. This is too much for a single PR. If you make a PR adding the index to the TransformResult table we will get a big speed up then we can take our time to discuss these changes and make sure we keep unit tests up to date.

Please re-work this as an incremental change. We can discuss the broader architectural changes over time

BenGalewsky (changes requested):
Too big for a single PR. Please make a PR for just adding the index and we should as a team discuss big architectural changes

docs/development/architecture.md
```diff
@@ -1,9 +1,10 @@
 # ServiceX Architecture
 
 ## Frontend
+
  ServiceX presents a RESTful interface for submitting transformation requests and
- requesting status updates. Most users will not wish to interact directly with 
- the REST interface so typically, ServiceX is accessed via clients distributed 
+ requesting status updates. Most users will not wish to interact directly with
```
BenGalewsky, line 6: Please be careful about introducing whitespace changes. It makes it hard to see what changed here...

servicex_app/servicex/lookup_result_processor.py
```diff
@@ -35,7 +36,10 @@ def __init__(self, rabbitmq_adaptor, advertised_endpoint):
 
     def add_file_to_dataset(self, submitted_request, dataset_file):
         request_id = submitted_request.request_id
-        dataset_file.save_to_db()
+
+        # dataset_file.save_to_db()
```
BenGalewsky, line 40: Remove rather then comment out code

servicex_app/servicex/models.py
```diff
@@ -215,6 +218,24 @@ def lookup(cls, key: Union[str, int]) -> Optional['TransformRequest']:
...
+        db.session.commit()
+
+    @classmethod
+    def file_transformed_unsuccessfully(cls, key: Union[str, int]) -> None:
+        req = cls.query.filter_by(request_id=key).one()
+        req.files_failed = cls.files_failed + 1
+        db.session.commit()
+
+    @classmethod
+    def add_a_file(cls, key) -> None:
```
BenGalewsky, line 234: Do you know what happens if another process updates this record while we are doing this? I assume the commit fails. How do we make sure we don't lose updates?

servicex_app/tests/resources/internal/test_transform_file_complete.py
```diff
@@ -25,8 +25,8 @@
 # CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-from servicex.models import DatasetFile, TransformationResult, TransformRequest
-from servicex.transformer_manager import TransformerManager
+from servicex.models import DatasetFile  # , TransformationResult, TransformRequest
+# from servicex.transformer_manager import TransformerManager
```
BenGalewsky: Please don't comment out code.

servicex_app/tests/resources/internal/test_transform_file_complete.py
```diff
@@ -43,97 +43,97 @@ def _generate_file_complete_request(self):
...
-        )
-
-        client = self._test_client(transformation_manager=mock_transformer_manager)
-        response = client.put('/servicex/internal/transformation/1234/file-complete',
-                              json=self._generate_file_complete_request())
-
-        assert response.status_code == 404
-        mock_transform_request_read.assert_called_with('1234')
-        mock_transformer_manager.shutdown_transformer_job.assert_not_called()
+    # def test_put_transform_file_complete_files_remaining(self, mocker):
```
BenGalewsky: Please don't comment out code

BenGalewsky (changes requested):
Also it looks like code coverage on servicex_app/servicex/resources/internal/transformer_file_complete.py dropped by 46%

https://app.codecov.io/gh/ssl-hep/ServiceX/pull/495

The reporting on codecov is a bit messed up, but the coverage percentages appears to be correct. Did you delete a bunch of tests there?

servicex_app/servicex/models.py
ivukotic: commit always does flush first, so we never lose updates.

servicex_app/servicex/lookup_result_processor.py
ivukotic: cleaned

docs/development/architecture.md
ivukotic: my VSCode enforces rules on md files too. I find the changes it makes or proposes usually improve the doc readability. But it gets too pedantic sometimes.

Issue #462, opened 2022-10-04
Use Official Docker Build/Push Action


BenGalewsky:
_As a ServiceX Developer I want the CI Job to use the official docker build/push action to make my builds faster and more reliable_

## Assumptions
1. Update the docker build actions to use the official [docker/build-push action](https://github.com/docker/build-push-action)
2. Set `cache-from` and `cache-to` properties so the builds can re-use layers from previous successful builds

BenGalewsky:
The clowder GHA already does this very well
https://github.com/clowder-framework/clowder/blob/develop/.github/workflows/docker.yml

BenGalewsky:
@shriram192 - it looks like we are cache thrashing and not getting any performance boost.
Please see if you could extend the matrix from `deploy-config.json` to signal whether to use the cache or not. We could reserve the cache for the really big images where we are spending most of the build time.

I see that the time for the xAOD Transformer is the same with and without the cache...

shriram192:
I have made a new branch called cache-ci, trying to utilize dockerhub registry for buildcache, it is working for those images, for images utilizing gha actions, i am unable to still see any caching, would suggest to use dockerhub registry for all the caching. Please have a look and let me know.

PR #498, opened 2022-11-09
Use Official Docker Build/Push Action #462

shriram192:


.github/FUNDING.yml
```diff
@@ -8,5 +8,5 @@ tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/
 community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
 liberapay: # Replace with a single Liberapay username
 issuehunt: # Replace with a single IssueHunt username
-otechie: # Replace with a single Otechie username
```
BenGalewsky: What is this?

BenGalewsky (changes requested):
Fix the odd addition to FUNDING.yaml and this is good to go! Really a nice improvement. Thanks for the research in finding this solution.

.github/FUNDING.yml
shriram192: I have reverted it, resolving the thread

BenGalewsky (approved):

Issue #483, opened 2022-10-27
Add Helm Lint to Helm Chart Release


BenGalewsky:
Re-enable helm lint and fix any errors that show up to get it running again

PR #499, opened 2022-11-12
Add Helm Lint to Helm Chart Release #483

prajwalkkumar:


BenGalewsky (approved):
Well, that was straightforward!

PR #504, opened 2022-11-21
Code Generator Service Returns Default Transformer Image #338

### LINKED ISSUE #338: Code Generator Service Returns Default Transformer Image
_As an Analyzer I want the code generator to be able to specify the transformer name and tag so I don't have to configure it in the app_

Part of Epic #309 

# Approach
Extend the response from the CodeGenerator services to include a string containing the image name and tag for the default transformer image. This is just a stepping stone on the way to Epic #309. We are moving the config from the app into the code generator and having the code generator report it back to the app.

# Assumptions
1. The values `transformer.defaultTransformerImage` and `transformer.defaultTransformerTag` move to `codeGen.defaultTransformerImage` and `codeGen.defaultTransformerTag` respectively. 
2. The code generator library is updated so that the [response from the POST operation](https://github.com/ssl-hep/ServiceX_Code_Generator_lib/blob/a67b3cc4fa9888ee60c25a25d1d2ae1484cb8420/servicex_codegen/post_operation.py#L91) becomes a multipart MIME type instead of just the octet stream. The result will include a JSON document containing the default transformer image and tag.
3. The App's [CodeGenaratorAdaptor](https://github.com/ssl-hep/ServiceX/blob/a291830a90fee7f92fcd5678f8cd2cd645f93bc1/servicex_app/servicex/code_gen_adapter.py#L58) will be updated to receive this from the result.
4. The transformer image can still be provided by the user, but it will have to be updated at code-gen time instead of at [submit time](https://github.com/ssl-hep/ServiceX/blob/a291830a90fee7f92fcd5678f8cd2cd645f93bc1/servicex_app/servicex/resources/transformation/submit.py#L81)

# Acceptance Criteria
Given I have a serviceX deployment with the `codeGen.defaultTransformerImage` value set and not `transformer.defaultTransformerImage` when I submit a transform request with no `image` property Then the transformer pods should start and be running the image specified in `codeGen.defaultTransformerImage`

Given I have a serviceX deployment when I submit a transform request with a value in the `image` property Then the transformer pods should start and be running the image specified in request

shriram192:


examples/ATLAS/.python-version
```diff
@@ -0,0 +1 @@
+3.7.14
```
BenGalewsky: I don't think you meant to check in your serviceX cache!

BenGalewsky: Actually, not sure what this file is

helm/servicex/templates/app/configmap.yaml
```diff
@@ -98,7 +98,7 @@ data:
     TRANSFORMER_SIDECAR_IMAGE = "{{ .Values.transformer.sidecarImage }}:{{ .Values.transformer.sidecarTag }}"
     TRANSFORMER_SIDECAR_PULL_POLICY = '{{ .Values.transformer.sidecarPullPolicy }}'
 
-    TRANSFORMER_SCIENCE_IMAGE = "{{ .Values.transformer.defaultScienceContainerImage }}:{{ .Values.transformer.defaultScienceContainerTag }}"
+    TRANSFORMER_SCIENCE_IMAGE = "{{ .Values.codeGen.defaultScienceContainerImage }}:{{ .Values.codeGen.defaultScienceContainerTag }}"
```
BenGalewsky, line 101: You will need to update the [helm publish](https://github.com/ssl-hep/ServiceX/blob/b9a6622cc32291f04e0145f49fded534e7476eb4/.github/workflows/ci_helm_publish.yaml#L55) GitHub action workflow

servicex_app/servicex/code_gen_adapter.py
```diff
@@ -36,8 +36,7 @@ def __init__(self, code_gen_url, transformer_manager):
         self.transformer_manager = transformer_manager
 
     def generate_code_for_selection(
-        self, request_record: TransformRequest, namespace: str
-    ):
+            self, request_record: TransformRequest, namespace: str, transformer_image: str = None):
         """
```
BenGalewsky, line 40: Need to update comment to reflect new argument

servicex_app/servicex/code_gen_adapter.py
```diff
@@ -48,16 +47,44 @@ def generate_code_for_selection(
...
         result = requests.post(self.code_gen_url + "/servicex/generated-code",
-                               data=request_record.selection)
+                               json=postObj)
 
         if result.status_code != 200:
             msg = result.json()['Message']
             raise ValueError(f'Failed to generate translation code: {msg}')
 
-        zipfile = ZipFile(BytesIO(result.content))
+        print(result.data)
```
BenGalewsky: Don't leave debugging statement in code

servicex_app/servicex/code_gen_adapter.py
```diff
@@ -48,16 +47,44 @@ def generate_code_for_selection(
...
+        boundary = result.data[2:34].decode('utf-8')
```
BenGalewsky: What is going on here? What is this magic _2:34_ magic number?

servicex_app/tests/test_code_gen_adapter.py
```diff
@@ -39,15 +38,31 @@ def _generate_test_request(self):
...
+        service = CodeGenAdapter("http://foo.com", mock_transformer_manager)
+        mock_transformer_image = "testimage"
+        service.generate_code_for_selection(self._generate_test_request(), "servicex", mock_transformer_image)
+        mock_requests_post.assert_called()
+
+        mock_transformer_manager.create_configmap_from_zip.assert_called_with(mock_zip(),
+                                                                              "462-33",
+                                                                              "servicex")
+
+    def test_generate_code_for_selection_without_params(self, mocker):
```
BenGalewsky: We need tests for exercising the Multipart mime return (`transformer_image=None`)

BenGalewsky (changes requested):

examples/ATLAS/.python-version
shriram192: python version file was added due to pyenv, i have removed it

helm/servicex/templates/app/configmap.yaml
shriram192: I have changed the helm ci for the merge
The changes made are here
Before -
.transformer.defaultScienceContainerTag = strenv(DOCKER_TAG) | .transformer.defaultScienceContainerTag style="double" |

After -
.codeGen.defaultScienceContainerTag = strenv(DOCKER_TAG) | .codeGen.defaultScienceContainerTag style="double" |

servicex_app/servicex/code_gen_adapter.py
shriram192: Added new :param comment to reflect the new parameter in function

servicex_app/servicex/code_gen_adapter.py
shriram192: removed the print statement, I will remove debug statements in further commits.

servicex_app/servicex/code_gen_adapter.py
shriram192: I am using a python package called requests_toolbelt for managing the multipart mime encoding and decoding.
This package adds a pseudo-random sequence of characters before and after the body and also between each multipart mime field. 2:34 is where the first boundary found in all the requests, for decoding the multipart mime the decoder requires the boundary, hence the magic number is needed.

servicex_app/tests/test_code_gen_adapter.py
shriram192: why would multipart mime return none? I am a bit confused on that, of the two cases we discussed.
1. App sends the transformer image to codegen, then codegen will return that image.
2. If app does not send the transformer image, the codegen returns the default config image.

In either case transformer image will not be None

servicex_app/servicex/code_gen_adapter.py
BenGalewsky: Please investigate further. I can't believe this is the correct way to interpret a multipart-mime response

servicex_app/tests/test_code_gen_adapter.py
BenGalewsky: Ok - I see what this argument is now... is there a good reason to send it to the code generator? Couldn't we allow the code gen to send back something but ignore it and use this provided one entirely in the App?

Also, we don't have any tests of the multipart mime unpacking

examples/ATLAS/Z_ee.py
```diff
@@ -5,10 +5,11 @@
 dataset = "mc16_13TeV:DAOD_TOPQ1.25521529._000009.pool.root.1"
```
BenGalewsky: This file shouldn't be in this PR

servicex_app/servicex/code_gen_adapter.py
```diff
@@ -36,28 +36,52 @@ def __init__(self, code_gen_url, transformer_manager):
...
-        self, request_record: TransformRequest, namespace: str
-    ):
+            self, request_record: TransformRequest, namespace: str):
         """
         Generates the C++ code for a request's selection string.
         Places the results in a ConfigMap resource in the
         Starts a transformation request, deploys transformers, and updates record.
         :param request_record: A TransformationRequest.
         :param namespace: Namespace in which to place resulting ConfigMap.
+        :param transformer_image: Set optional transformer image name for the request
```
BenGalewsky: Remove this parameter comment

servicex_app/servicex/code_gen_adapter.py
```diff
@@ -36,28 +36,52 @@ def __init__(self, code_gen_url, transformer_manager):
...
+            zipfile = ZipFile(BytesIO(zipfile))
+
+            return (self.transformer_manager.create_configmap_from_zip(zipfile,
+                                                                       request_record.request_id,
+                                                                       namespace), transformer_image)
+        else:
+            zipfile = ZipFile(BytesIO(""))
+            return (self.transformer_manager.create_configmap_from_zip(zipfile,
+                                                                       request_record.request_id,
+                                                                       namespace), transformer_image)
```
BenGalewsky: I wonder if we should report an error if `decoder_parts.parts` <> 2

Certainly `transformer_image` is undefined here

BenGalewsky (changes requested):
Excellent! Congratulations on figuring out that test!
Couple of commit clean ups, and a bug in what happens if the code generator doesn't return the correct number of multipart mime parts....

servicex_app/servicex/code_gen_adapter.py
BenGalewsky: Let's not have print statements in production code...

examples/ATLAS/Z_ee.py
shriram192: The Z_ee.py? It was present in the develop branch, do you want it removed?

servicex_app/servicex/code_gen_adapter.py
shriram192: I had removed the print previously, it is not present in the code.

servicex_app/servicex/code_gen_adapter.py
shriram192: I don't think it would happen, I had just put up this condition for the test case to run, incase the codegen lib fails, it would return a code 500, otherwise there is no reason for it to return decoder_parts < 2

shriram192: I have removed the check for decoder_parts len < 1 please do check the new version, all test cases are working fine on  it.

servicex_app/servicex/code_gen_adapter.py
shriram192: it has been removed

BenGalewsky (approved):

Issue #500, opened 2022-11-17
CERN OpenData DID Finder Fails


BenGalewsky:
# Problem
Upon deploying the helm chart with the CERN open data DID finder enabled, the process is stuck in _CrashLoopBackoff_

# Approach
It looks like the CERN OpenData DID Finder comes up faster than all of the other services. It attempts to connect to the App, but it is not running. 

Use an init-container with the NCAS Checks docker image to wait for the app to come up before starting up the CERN OpenData did Finder

BenGalewsky:
If more than one point to fix then write a ticket for the next sprint

PR #508, opened 2022-11-29
#500 Add init container to CERN open data finder to wait for app to come up

prajwalkkumar:


BenGalewsky (approved):
Great - this is a nice fix

PR #509, opened 2022-11-29
Minio cleanup

sthapa:
This should have all the code needed to let ServiceX cleanup S3 data.  This avoids the situation where the S3 storage fills up and overflows and results in ServiceX giving odd errors.  This has occurred a few times at Chicago and UNL so it's something that happens somewhat frequently. There's two things that need to be handled a bit carefully, the minio cleanup image creation and naming.  I think I have that setup in the workflows.  Also, the code depends on an library for storage access, right now it's a github ref but getting it into pip would be good.


ivukotic (commented):
I thought you will make it completely standalone?
Anyhow, I get a bit irked by all the Minio references when all of this will work with any s3 compatible storage. the only Minio specific thing is a storage we deploy if user does not provide its own s3 storage.

ivukotic (commented):
Can you get this logging via logstash too?

ivukotic (commented):
Have you tested it?
From this PR I don't see the part that actually does cleanup?

ivukotic (commented):
I don't see that the documentation has been updated? I have no idea what two of the variables mean.

sthapa:
@ivukotic Addressing your comments in order:

* [This file](https://github.com/ssl-hep/ServiceX/blob/minio_cleanup/minio_cleanup/resources/cleanup_batch.yaml) has a manifest that can be used to deploy the cronjob manually.  Also I left the minio names in the parameters to match what we have, we can change the chart but we should do that globally instead of piecemeal.  The code itself calls it s3 rather than minio
* Sure I can add logstash formatting
* I tested this locally referring to a locally built image (using the project dockerfile) and verifying that the cronjob would remove files as needed.  Most of the work actually occurs in the [module](https://github.com/ssl-hep/servicex_storage_lib)
* Added documentation for the chart just now

ivukotic:
> 

why would you want it deployed manually and not automatically with servicex?

sthapa:
> > 
> 
> why would you want it deployed manually and not automatically with servicex?

If people use the helm chart, they can automatically deploy this with servicex.  But if a site is using a standalone storage instance, they may want the cronjob that cleans it up running in the same location (cluster, etc.).

sthapa:
@ivukotic added in the logstash logging so everything should be good to go now

ivukotic (approved):
LGTM

PR #519, opened 2022-12-08
Uproot transformer writes root files

BenGalewsky:


code_generator_funcadl_uproot/servicex/templates/transform_single_file.py
```diff
@@ -25,22 +25,29 @@ def transform_single_file(file_path: str, output_path: Path):
...
-        explode_records = bool(awkward_array.fields)
-        try:
-            arrow = ak.to_arrow_table(awkward_array, explode_records=explode_records)
-        except TypeError:
-            arrow = ak.to_arrow_table(ak.repartition(awkward_array, None),
-                                      explode_records=explode_records)
+        if output_format == 'root-file':
+            etime = time.time()
+            with uproot.recreate(output_path) as writer:
+                writer[default_tree_name] = awkward_array
```
masonproffitt: I think we can work around https://github.com/scikit-hep/uproot5/issues/777 by using something like:
```python
writer[default_tree_name] = {field: awkward_array[field] for field in awkward_array.fields} if awkward_array.fields else awkward_array
```

PR #527, opened 2022-12-22
Added CLI support to add users.

prajwalkkumar:


servicex_app/servicex/models.py
```diff
@@ -49,7 +49,7 @@ class UserModel(db.Model):
     institution = db.Column(db.String(120))
     name = db.Column(db.String(120), nullable=False)
     pending = db.Column(db.Boolean, default=True)
-    refresh_token = db.Column(db.Text, nullable=False, unique=True)
+    refresh_token = db.Column(db.Text, nullable=True, unique=True)
```
BenGalewsky, line 52: Need to add a migration for this

servicex_app/servicex/cli/create_default_user.py
```diff
@@ -0,0 +1,25 @@
+from kubernetes import client, config
```
BenGalewsky: Why doesn't this use the mounted secret volume?

helm/servicex/templates/app/deployment.yaml
```diff
@@ -130,8 +130,13 @@ spec:
             mountPath: /opt/servicex
           - name: sqlite
             mountPath: /sqlite
+          {{- if .Values.defaultUsers }}
```
BenGalewsky: This should be app.defaultUsers, no?

BenGalewsky (approved):
This is great! Thank you

PR #531, opened 2023-01-10
Add RABBIT_MQ URI as a value in helm values

prajwalkkumar:


BenGalewsky:
Good start!
We need to make sure the default chart works as before (easily deploy with integrated RabbitMQ sub chart) - I think this also needs some [documentation updates](https://github.com/ssl-hep/ServiceX/tree/develop/docs/deployment)

prajwalkkumar:
Updated documentation and added default value to values.yaml

BenGalewsky (changes requested):
This doesn't work. You can see with the default values.yaml
```
helm template -f local-values/local-sidecar-values.yaml servicex-ben servicex | grep RABBIT_URI -C 2
```
That RABBIT_URI is blank. It looks like you need some extra stuff to be able to use template functions inside values.yaml

See this [StackOverflow answer](https://stackoverflow.com/a/62896363)

prajwalkkumar:
Any reason why RABBIT_URI is outside the conditional? I added it inside as the variable was used only inside the conditional in all deployments.

ponyisi:
I'm assuming we can close this?

ponyisi:
Closing; if we need to use an external rabbit service we can try this again.

PR #550, opened 2023-02-27
Xcache support 4 cernopendata

ivukotic:
just adds a prefix if it was defined.


did_finder_cernopendata/src/servicex_did_finder_cernopendata/did_finder.py
```diff
@@ -46,11 +49,11 @@ async def find_files(did_name: str,
             assert isinstance(line, str)
             all_lines.append(line)
             uri = line.strip()
-            if not uri.startswith('root://'):
+            if uri.startswith('http://') and cache_prefix == '':
```
gordonwatts, line 52: Is there a case where it returns `http` - and with the cache prefix that makes it ok?

gordonwatts (approved):
This looks great!

PR #551, opened 2023-02-28
Force restart of app Pod when configuration changes.

### LINKED ISSUE #297: ServiceX_App doesn't recognize when its config information is updated in the helm chart with helm upgrade
**Describe the bug**
If you upgrade your charts values for the ServiceX_App (e.g. change something like the default did finder), and then run `helm upgrade` the command will succeed as expected. However, `ServiceX_App` will not recognize the new values.

**To Reproduce**
1. Run a services chart.
2. Alter a value that is aimed at the `ServiceX_App` flask container
3. Use `helm upgrade` to update the running system
4. Note that the `ServiceX_App` does not restart, and thus does not re-read the settings in the config map that contains these options.

**Expected behavior**
`ServiceX_App` should be aware of the new values, or be forced to restart (thus making it aware).

draghuram:
Fixes #297.

I tested the fix as follows:

- Installed chart from local directory.
- Changed "objectStore.publicURL", did ``helm upgrade``, and verified that app pod didn't restart. 

I repeated above steps with the fix and I could see that app pod is restarted.

Thanks to @AndrewEckart for his suggestion in #297.

### DIFF
--- a/helm/servicex/templates/app/deployment.yaml
+++ b/helm/servicex/templates/app/deployment.yaml
@@ -12,6 +12,8 @@ spec:
       labels:
         app: {{ .Release.Name }}-servicex-app
         helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/app/configmap.yaml") . | sha256sum }}
     spec:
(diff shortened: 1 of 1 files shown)


BenGalewsky (approved):
Thank you so much for your contribution! This works a charm

draghuram:
No problem. I am happy to see the PR merged.

Issue #539, opened 2023-02-10
ServiceX DID-finder stalls for Rucio dataset accessed via https protocol


kyungeonchoi:
**Describe the bug**
ServiceX DID-finder stalls for Rucio dataset accessed via https protocol. A ServiceX request seems to wait for the file list returned from DID-finder. Once Rucio dataset is moved to MWT2 (xrood protocol) ServiceX can get the file list and runs fine. 

An example Rucio RSE for https protocol: 
`NEVIS_GRIDFTP: https://xenia.nevis.columbia.edu:1094/atlas/dq2/rucio/user/dpanchal/be/a4/user.dpanchal.31107845._000054.hist-output.root`

ivukotic:
Well, I would not call it a bug... All of the ATLAS storages are required to provide an xroot endpoint. So this is an issue of the Nevis storage. On the other hand we would like to be able to use (and also cache) data read over http. I am working on it, and if/when performance is good enough, I will make it possible to use both xroot and http.

ponyisi:
Hi Ilija - understood that it's an issue with Nevis not necessarily supporting the "proper" endpoint, but I consider it a problem if we cannot access a file that is available via "rucio get" (as this one is). KyungEon, you reported a site that was webdav ?

kyungeonchoi:
Thanks @ivukotic for looking into this problem. I didn't know that http doesn't need to be a supported endpoint. As you said it would be still good to support both xrootd and http.
@ponyisi - I will talk to Dev.

ivukotic:
Hi,

I just started work on getting xcaches to be able to use paths like root://xcache.xx.edu//http://origin.xx.edu/…

Ilija

PR #553, opened 2023-03-02
moving prefix handling from finders to transformer

ivukotic:
We want to be able to use http protocol to access data inaccessible via xroot. 
(https://github.com/ssl-hep/ServiceX/issues/539)
Now XCache supports paths like root://xcache.xxx.org/http://origin/path.df
We also want to use xcache in situations which don't involve did finding (user gives single paths).

Solution is to prepend path in transformer sidecar and not in did finders. 
http:// paths will be tried only after root:// paths are exhausted. 

This change involves changes in chart, did finders, side car, documentation.


servicex_app/servicex/transformer_manager.py
```diff
@@ -113,20 +113,17 @@ def create_job_object(request_id, image, rabbitmq_uri, workers,
                 field_path="metadata.name"))
         env += [client.V1EnvVar("POD_NAME", value_from=pod_name_value_from)]
 
-        # provide pods with level and logging server info
```
BenGalewsky, line 116: Where does this come from now?

ivukotic: for some reason this part of code was duplicated. I just removed one copy of it.

BenGalewsky (approved):
This is such a great idea -- Approved!

PR #566, opened 2023-03-30
Fix python codegen to work with sidecar

BenGalewsky:
It was missing some of the extra files required for the sidecar transformer to work


prajwalkkumar:
@BenGalewsky I was taking a look/testing  this. The "transformer_capabilities.json" file is not added to the image in the  Dockerfile. I think we need that step after this change https://github.com/ssl-hep/ServiceX_Code_Generator_lib/commit/e551f6cf2e3e2516f4ea5f21f13206fb40ae503f

PR #570, opened 2023-04-12
fix: Set workflow output parameters with GITHUB_OUTPUT

matthewfeickert:
* Update to use the new GitHub workflow GITHUB_OUTPUT environmental file to avoid problems when the setting 'set-output' via stdout is deprecated in 2023. c.f.:
   - https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
   - https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files
   - https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-setting-an-output-parameter

### DIFF
--- a/.github/workflows/ci_servicex.yaml
+++ b/.github/workflows/ci_servicex.yaml
@@ -27,7 +27,7 @@ jobs:
           content="${content//$'\n'/'%0A'}"
           content="${content//$'\r'/'%0D'}"
           # end of optional handling for multi line json
-          echo "::set-output name=matrix::$content"
+          echo "matrix=$content" >> $GITHUB_OUTPUT
       - run: |
           echo "${{ steps.set-matrix.outputs.matrix }}"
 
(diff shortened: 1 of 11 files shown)


.github/workflows/ci_servicex.yaml
```diff
@@ -85,12 +85,12 @@ jobs:
       - name: Extract tag name
         working-directory: ${{ matrix.app.dir_name }}
         shell: bash
-        run: echo "##[set-output name=imagetag;]$(echo sslhep/${{matrix.app.image_name}}:${GITHUB_REF##*/})"
+        run: echo "imagetag=$(echo sslhep/${{ matrix.app.image_name }}:${GITHUB_REF##*/})" >> $GITHUB_OUTPUT
```
matthewfeickert, line 88: In the current existing

```yaml
      run: echo "##[set-output name=imagetag;]$(echo sslhep/${{matrix.app.image_name}}:${GITHUB_REF##*/})"
```

I'm not very clear on what the first `##` was intended to strip out, so this subset of changes might be missing the original point of things like https://github.com/ssl-hep/ServiceX/commit/9f269882ba4e4e22ab3aa239aa4d10973a26d2c1.

matthewfeickert:
This is ready for review.

.github/workflows/ci_servicex.yaml
ivukotic: I just found this PR. It should be encompassed in the one I made: 
https://github.com/ssl-hep/ServiceX/pull/603
Please feel free to review...

BenGalewsky:
Superseded by #603

PR #574, opened 2023-05-01
Unzip Transformer, DID Finder, Pandas Python Code Generators

### LINKED ISSUE #341: Unzip Transformer
# Story
_As an Astrophysicist, I want to extract compressed data from the Spitzer Photometry and Accurate Rotation Curves website so I can get accepted rotation curves for my analysis_

# Description
There are zip files available to download via http. We would like to perform analysis on the data contained within them so we need a way to inject them into the ServiceX object store.

This is the first case of a transformer that doesn't require a code-gen. We will use this opportunity to make the code-gen property of the transform request optional. If the code-gen property is not included, the transform workflow will skip the code generation step. The transformer manager will handle the case that there is no generated code configmap to mount in the transformers.

The user will provide the transformer "science container" in the transform request. It will be `sslhep/unzip-transformer:latest`

The transformer will download the zip file via HTTP and unzip the results into the usual output minio bucket.

See if we can use a [streaming unzip](https://stream-unzip.docs.data.trade.gov.uk) to make this scalable by avoiding keeping the entire zip file and unzipped contents local to the transformer.

# Test Case
We specifically want to see how this works with http://astroweb.cwru.edu/SPARC/Rotmod_LTG.zip

### LINKED ISSUE #342: ServiceX Transform DID Finder
# Story
_As an analyzer I want to chain transforms together so I can perform more complicated transforms_

# Assumptions
1. The "DID" is the transform ID of a previously submitted request
2. Objects from the minio bucket are streamed out as the results
3. The DID Finder waits til the source transform has completed (successfully or not) before signaling the end of the dataset

### LINKED ISSUE #562: Pandas Python Code Generator
# Story
_As a analyzer I want to be able to work with data in CSV files so I can standardize my analysis_

# Description
We want to be able to read data from weird CSV and TSV formats from an experiment. We will use the Pandas library as the basis for this transformer.

We will use the python code generator with the following assumptions:
1. The function will be called with an open file handle
2. The function will return a Pandas dataframe

There will be a new python code generator based on the [uproot python code generator](https://github.com/ssl-hep/ServiceX/tree/develop/code_generator_python).  Its `transform_single_file.py` script will write the data frame to parquet using   the dataframe [to_parquet method](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_parquet.html).

It looks like it is possible to stream directly from the Dataframe to a parquet object in minio https://stackoverflow.com/a/57838851

prajwalkkumar:


BenGalewsky:
Put pandas lib into the python code gen image
The S3 DID Finder should be pulled out into a separate PR
The Unzip transformer will not be used

PR #582, opened 2023-05-12
reduced sleeps

### LINKED ISSUE #581: sleep times in sidecar are a significant slow down for transformers processing small files
Most of the files we process are in 100 MB range and frequently access only a tiny part of it.
So transformer container often processes single file in <2 seconds. 
Sidecar has sleep of 1 second. This should be reduced.

ivukotic:


BenGalewsky (approved):
Thank you so much for spotting this. Approved!

PR #583, opened 2023-05-15
now returning correct servicex.yaml

ivukotic:


BenGalewsky (approved):
This is a good solution to a limitation in the client code

PR #595, opened 2023-05-26
Python transformer for multi-tree ROOT output - bugfix

kyungeonchoi:
- Fix bug when `output_format` is set to parquet


kyungeonchoi:
@BenGalewsky - Unfortunately I found a bug when `output_format` is set to parquet. This change fixes that bug.

BenGalewsky (approved):
More complete now. Thanks

PR #601, opened 2023-06-26
597 make sidecar and transformer containers communicate over socket

### LINKED ISSUE #597: Make sidecar and transformer containers communicate over socket
On sidecar side use regular python and on transformer side use netcat or just bash to let sidecar know the processing has been done.

ivukotic:
This PR makes sidecar - transformer communication go over a localhost socket. 
That is more reliable and performant than communication over files in a shared directory. Upload to s3/minio is now correctly separated from the transformation loop. 
Code is much simpler.

It does require a new version of science images coming from https://github.com/ssl-hep/science-images/pull/1


BenGalewsky (approved):
Ok - this looks like a big improvement - I removed some commented out code. I do wonder about some of the print statements still in the code....

PR #603, opened 2023-06-28
602 GitHub actions need an update

### LINKED ISSUE #602: Github actions need an update
There are a lot or warnings concerning using set-output. 
There are actions that use "@master" instead of version and these are years old.
One of the actions was archived two years ago.

ivukotic:
fixes all the warnings.
updates to new versions of all actions.


BenGalewsky (approved):
Great update! Thank you

PR #605, opened 2023-07-05
better replica filtering

### LINKED ISSUE #604: Rucio did finder replica filtering
Rucio client by default returns all the replicas, even the ones on RSEs currently offline. 
It also returns replicas on Tape RSEs.
There is now a better way to get location.

ivukotic:


BenGalewsky (approved):
@ivukotic thank you for bringing your Rucio experience to this service!

PR #622, opened 2023-08-04
621: Only cast LOGSTASH_PORT if LOGSTASH_HOST is set

### LINKED ISSUE #621: Transformer crashes if logging disabled
If you set `logging.logstash.enabled` to `false`, the transformer crashes.

This is because `LOGSTASH_PORT` and `LOGSTASH_HOST` do not get set in the app env, but:  https://github.com/ssl-hep/ServiceX/blob/develop/servicex_app/servicex/transformer_manager.py#L104-L108 sets it to none, and https://github.com/ssl-hep/ServiceX/blob/develop/transformer_sidecar/src/transformer_sidecar/transformer_logging/__init__.py#L58 casts to int.

holzman:
Fixes #621


ivukotic (approved):
LGTM

PR #628, opened 2023-08-17
627: Explicitly set POETRY_CACHE_DIR to enable running as user

### LINKED ISSUE #627: did_finder_cernopendata needs POETRY_CACHE_DIR explicitly set on OpenShift
On OpenShift, the did_finder runs as an ephemeral UID with `HOME=/` and the following error occurs:

```
virtualenv: error: argument dest: the destination . is not write-able at /
```

The software is installed as root and into the default `poetry.cache-dir`, which is `$HOME/.cache/pypoetry`.
We just need to set `POETRY_CACHE_DIR=/root/.cache/pypoetry` in the image or as an env in the helm chart.

holzman:
Fixes #627


BenGalewsky (approved):
Sweet and simple. Thanks

PR #629, opened 2023-08-30
getting correct IPv6 vs IPv4 preferences

ivukotic:
By default, k8s deployments will preferentially use IPv4. This PR will make it prefer IPv6, just like a regular OS.

Preference is controlled by /etc/gai.conf file. On regular OS IPv6 is preferred since it is more performant. But on stock kubernetes clusters default is IPv4. 

This change will not affect performance on single stack IPv4 clusters but will improve it on dual stack ones.

### DIFF
--- a/servicex_app/Dockerfile
+++ b/servicex_app/Dockerfile
@@ -17,6 +17,7 @@ RUN poetry config virtualenvs.create false && \
 COPY *.py docker-dev.conf boot.sh ./
 COPY servicex/ ./servicex
 COPY migrations migrations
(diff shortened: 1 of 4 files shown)


BenGalewsky:
Could you update the description of this PR to explain a little more how this works...
Also, what will the impact be on IPv4 networks?

ivukotic:
Just edited description. The change will not affect IPv4 performance.

BenGalewsky (approved):
No idea what this is doing, but trust it will help!

PR #630, opened 2023-08-30
Move migrate to init and remove redundant db.init

prajwalkkumar:
Previously we initialised the db twice, once inside create_app() and once in app.py, and SQLAlchemy/Flask did not seem to mind. Now, we are not allowed to initalise the db more than once per flask app.  We now initialise the db and perform migrations (without the db.create_all() call, as the migrations take care of the create_all() function call) for the happy case.


BenGalewsky (approved):
Deceptively simple! Thank you

PR #638, opened 2023-09-18
496 reorganize dataset and file handling in the db

### LINKED ISSUE #496: Reorganize dataset and file handling in the DB
Since we already have postgresql we should use it to store file information and not rely on external memcache. 

step 1.
add "dataset" table
* ds_id
* last_used
* last_updated
* origin (eg. rucio)
* nfiles
* size
* events
* replicas

step 2.
current "files" table should be changed: 
* replace request_id with  ds_id

step 3. 
in table requests, replace did with ds_id foreign key.

step 4.
add code that periodically goes through all the datasets that were updated more than 6h ago and if needed updates files related to them. It goes through all the datasets older than 10 days and removes them and all the files related.

step 5. 
add code that before calling did-finder, checks if dataset is already registered and if yes uses existing data from files table.

ivukotic:
This very large PR makes an important change in DB:
* it adds a Dataset table. Request points to a dataset in that table. Dataset table points to files in the Datafile table.
* Now a request for a set that was already seen does not contact did finder at all and starts right away. It also keeps datafile table to a reasonable size. 
*  DID finder now needs to know dataset_id, so that makes a breaking change in did lib that needs to be released as v2.0.0
* chart also removes memcache from did finder as it is not needed anymore.
* it removed python 3.7 from rucio did finder as now did lib does not support 3.7
* it updated all github actions. 
* removed num-messages from side-container as it is not in db anymore. 
* it updates all the affected tests.
* it updated documentation


servicex_app/migrations/versions/v1_3_0.py
```diff
@@ -0,0 +1,123 @@
...
+
+    # this should have worked...
+    # sa.Column('dataset_id', sa.Integer(), key=sa.ForeignKey(
+    # "datasets.id", ondelete='cascade'), nullable=False),
+
+    op.create_index(op.f('ix_datasets_id'),
+                    'files', ['dataset_id'], unique=False)
+
+    #   Change requests table
+    op.add_column('requests', sa.Column('did_id', sa.Integer(), nullable=False))
```
BenGalewsky, line 70: Why isn't this column called dataset_id? That would be more consistent

BenGalewsky:
Ok - I get this now. It's a great improvement.

My biggest concern right now is what happens if someone submits requests with the same DID at the same time? There might be logic in there to handle it (some comments would be appreciated),  however it could be greatly simplified if we said that the DID Finders only care about `dataset_id` and don't even know about transform requests. 

Then the code that handles the `add_files` operation could query the db to find all of the requests that depend on that dataset_id and put the files in their queues.

Also, what about flushing datasets out of the cache? I assume DIDs could change over time. Ideally there should be a cron job that flushes old datasets. We'd also want a way for a user to say for a particular request that it should ignore the cache and fetch fresh files from the DID Finder.

servicex_app/migrations/versions/v1_3_0.py
ivukotic: will rename it now.

ivukotic:
> My biggest concern right now is what happens if someone submits requests with the same DID at the same time? There might be logic in there to handle it (some comments would be appreciated), 

In this case the first request will create a dataset but that dataset won't be in "complete", so it won't start transformers. Only when in complete transformers start (either from submit.py or from did finder calling start endpoint).

> however it could be greatly simplified if we said that the DID Finders only care about `dataset_id` and don't even know about transform requests. Then the code that handles the `add_files` operation could query the db to find all of the requests that depend on that dataset_id and put the files in their queues.

that was how it was done originally, but Gordon wanted the submit code to be asynchronous so I've put back "start" endpoint to be called by did finder on completion.

> Also, what about flushing datasets out of the cache? I assume DIDs could change over time. Ideally there should be a cron job that flushes old datasets. We'd also want a way for a user to say for a particular request that it should ignore the cache and fetch fresh files from the DID Finder.

DIDs can't change. Or the files they contain. The thing that can change are the file paths. For that reason datasets have fields: last_used and last_updated. I will delete from the db datasets not used in last 3 months and will refetch  the rest if older than 3 days. I had that part written but can't find it now. Will do it again this week. But let's not wait for it ...

BenGalewsky:
> DIDs can't change. Or the files they contain. The thing that can change are the file paths. For that reason datasets have fields: last_used and last_updated. I will delete from the db datasets not used in last 3 months and will refetch the rest if older than 3 days.

Please write an issue for this

BenGalewsky:
> that was how it was done originally, but Gordon wanted the submit code to be asynchronous so I've put back "start" endpoint to be called by did finder on completion.

I don't understand why this would not be asynchronous. Wouldn't the first `AddFileToDataset` be the trigger to start all of the transforms dependent on that dataset_id?

Also I don't see how the `start` endpoint would pick up the second transform request. It just uses the request_id coming back from the did finder.

I would love to see a unit test that demonstrates that the start works correctly where there are multiple transforms waiting on this dataset

ivukotic:
closing this one as the squashed one is open

PR #672, opened 2023-11-21
Pr_xAOD_codegen

### LINKED ISSUE #671: Add list of code generators to the about page
# Story
_As a ServiceX user I want to quickly see what code generators are available g so I can know what code generator arguments to use in my queries_

# Other Methods

1. You can currently use the CLI to do this, but that means you have to have this installed - sometimes it is just easier to load a web page.

# Acceptance Criteria
1. Load the about page for a servicex instance and see what is up there.

gordonwatts:
* Update the default app_config file to contain a few more things required in the `about.html`
* Modify `about.html` to include references to the code-get images.

Fixes #671

### DIFF
--- a/servicex_app/servicex/templates/about.html
+++ b/servicex_app/servicex/templates/about.html
@@ -25,6 +25,12 @@
         <td>DID Finder CERN Open Data</td>
         <td>{{ config['DID_CERNOPENDATA_FINDER_TAG']}}</td>
       </tr>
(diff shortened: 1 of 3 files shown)


gordonwatts:
The test failure seems to be a pipeline problem, not something to do with this code. :-/

BenGalewsky (approved):
Looks like a good addition

PR #691, opened 2024-01-17
Update parquet serialization to be compatible with awkward2

kyungeonchoi:
Update parquet serialization in the Python transformer to be compatible with awkward2


BenGalewsky (approved):
Thanks for figuring this out

PR #692, opened 2024-01-18
Fix parquet serialization in Python transformer

kyungeonchoi:
Fix parquet serialization in Python transformer


BenGalewsky (approved):
I like the way this is looking!

PR #695, opened 2024-02-09
Move uproot code generator to a uniquely named package

BenGalewsky:
# Problem
Many of the services in the monoRepo use the same package name, `servicex` - this causes confusion with IDEs and code coverage reports.

This PR is one of many to migrate the services to unique package names


ivukotic (commented):
works with my fix.

PR #700, opened 2024-02-13
Update GitHub Actions versions

ivukotic:
the old ones where generating warnings due to their usage of old node js


BenGalewsky (approved):
Thanks for updating this

PR #720, opened 2024-03-26
Fixing logging

ivukotic:
fixes number of issues where logger was not reporting host_name and site. 
It also changed all the request_id and request-id to requestId 
Unrelated to logging it fixed the chart where endpoint was not used in ADVERTIZED_HOST.

If not too much work it would be nice to have this as 1.3.3

### DIFF
--- a/servicex_app/tests/resources/transformation/test_submit.py
+++ b/servicex_app/tests/resources/transformation/test_submit.py
@@ -87,7 +87,8 @@ class TestSubmitTransformationRequest(ResourceTestBase):
         dm.name = hash
         dm.id = 43
 
-        mock_from_file_list = mocker.patch.object(DatasetManager, "from_file_list", return_value=dm)
+        mock_from_file_list = mocker.patch.object(
+            DatasetManager, "from_file_list", return_value=dm)
         return mock_from_file_list
 
     @fixture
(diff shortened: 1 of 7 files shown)


gordonwatts:
Other than the one comment and the two minor ones about a short comment so the reader knows what things are, this looks good. Hopefully @BenGalewsky can approve fairly quickly when he is back tomorrow.

I will note that the extras wiht logging infomration are now getting quite complex - so just logging is taking up lots and lots of lines of code. I can't help but wonder if there isn't a better way to do this in the future.

Like someone wrote some code that is a decorator and any logging called inside a routine with that decorator would automatically add the notation (I'm just guessing here).

servicex_app/tests/resources/transformation/test_submit.py
```diff
@@ -210,7 +212,8 @@ def test_submit_transformation_default_scheme(self, mock_rabbit_adaptor,
                                                   mock_dataset_manager_from_did,
                                                   mock_codegen,
                                                   mock_app_version):
-        client = self._test_client(rabbit_adaptor=mock_rabbit_adaptor, code_gen_service=mock_codegen)
```
BenGalewsky, line 213: In future, please avoid reformatting unrelated code - it makes the PR so much more noisy and hard to review

BenGalewsky (approved):
Please be kind to your reviewer if you want quick reviews. Be careful to not reformat unrelated code. It makes the PRs noisy and hard to see what changed

PR #722, opened 2024-04-05
simplify secret naming. 

ivukotic:
makes it possible to test servicexlite against testing instances.
there is really no reason to add instance name since helm deploys in different namespaces. other secrets don't have instance name prepended.


BenGalewsky (approved):
ok

PR #726, opened 2024-04-08
multi-xcache-support

ivukotic:
now one can specify more than one xcache in the same string, just have them comma separated.
Transformer sidecar will prepend one calculated based on filename hash.

### DIFF
--- a/transformer_sidecar/src/transformer_sidecar/transformer.py
+++ b/transformer_sidecar/src/transformer_sidecar/transformer.py
@@ -34,7 +34,7 @@ import os
 import psutil as psutil
 import shutil
 import timeit
-from hashlib import sha1
+from hashlib import sha1, sha256
 from pathlib import Path
 from queue import Queue
 from typing import NamedTuple
@@ -120,7 +120,23 @@ def fill_stats_parser(stats_parser_name: str, logfile_path: Path) -> Transformer
     return globals()[stats_parser_name](logfile_path)
 
 
+def prepend_xcache(file_paths):
+    # if CACHE_PREFIX is given, returns file paths prefixed with
+    # a cache determined by the filename. Cache list is comma separated.
+    prefix = os.environ.get('CACHE_PREFIX', '')
+    if not prefix:
+        return file_paths
+    xcs = [p.strip() for p in prefix.split(',')]
+    prefixed_paths = []
+    for f in file_paths:
+        hex_digest = sha256(f.encode()).hexdigest()
+        c = int(hex_digest, 16) % len(xcs)
+        prefixed_paths.append(f'root://{xcs[c]}//{f}')
+    return prefixed_paths
+
 # noinspection PyUnusedLocal
+
+
 def callback(channel, method, properties, body):
     """
     This is the main function for the transformer. It is called whenever a new message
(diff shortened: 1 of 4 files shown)


gordonwatts:
This looks fine to me - but really, @BenGalewsky needs to approve these. The only tihng is that the pre-pending code is starting to be come a little complex and has three cases it has to correctly deal with:

* ''
* a single item
* multiple items

And the latter needs to make sure it switches off on multiple files or a single file.. In short, it is starting to become complex enough that a test might be in order.

But I'll leave Ben to make the final decision.

transformer_sidecar/src/transformer_sidecar/transformer.py
```diff
@@ -120,7 +120,23 @@ def fill_stats_parser(stats_parser_name: str, logfile_path: Path) -> Transformer
...
+def prepend_xcache(file_paths):
+    # if CACHE_PREFIX is given, returns file paths prefixed with
+    # a cache determined by the filename. Cache list is comma separated.
+    prefix = os.environ.get('CACHE_PREFIX', '')
+    if not prefix:
+        return file_paths
+    xcs = [p.strip() for p in prefix.split(',')]
+    prefixed_paths = []
+    for f in file_paths:
+        hex_digest = sha256(f.encode()).hexdigest()
```
BenGalewsky, line 147: What is this magic here? The comment doesn't explain any of this

BenGalewsky (changes requested):
Please make the comments describe the magic that is happening with the hex_digest.

Can you confirm that this will still work with existing deployments? 

Also, it looks like we picked up a new version of the helm chart lint process and the CI job is failing. I'd appreciate you fixing the problem it is highlighting

transformer_sidecar/src/transformer_sidecar/transformer.py
```diff
@@ -120,7 +120,27 @@ def fill_stats_parser(stats_parser_name: str, logfile_path: Path) -> Transformer
     return globals()[stats_parser_name](logfile_path)
 
 
+def prepend_xcache(file_paths):
+    # if CACHE_PREFIX is given, returns file paths prefixed with
+    # a cache determined by the filename. Cache list is comma separated.
+    prefix = os.environ.get('CACHE_PREFIX', '')
+    if not prefix:
+        return file_paths
+    # split the string into a list of xcaches. strip in case someone adds spaces
```
BenGalewsky, line 140: This is describing the code, but I still have no idea how we get from the hash/hex digest to the Xchache. Maybe a through description in the ServiceX `values.yaml` file for the `cachePrefix` setting. It seems like there is some external configuration that this is referring to but it is not apparent to me

BenGalewsky (changes requested):
Please put somewhere a description of the implementation of the multiple xCahches that uses this code

transformer_sidecar/src/transformer_sidecar/transformer.py
ivukotic: I am not sure I understand your question... we now expect cachePrefix setting to be eg. "cache1.uc.edu" or "cache1.uc.edu,cache2.uc.edu,whatever.edu" this string will be split into a list of prefixes. only one of them will be prepended to the file path. Exactly which one depends on the hash of the filename.
Does this explain?

BenGalewsky (approved):
Great... thank you

PR #727, opened 2024-04-11
 Update func-adl-uproot to v2.1.1

masonproffitt:
This updates the FuncADL Uproot code generator to v2.1.1. This version brings in a significant fix that allows running on ROOT files with branches not intpretable by `uproot`/`awkward`, which includes DAOD_PHYSLITE files.


BenGalewsky (approved):
Thanks

PR #729, opened 2024-04-11
Clean up the prepend_xcache comments

BenGalewsky:
Since it took me so long to understand what was happening here, I wanted to make sure the comments were as complete as I could make them. Also make the variable names blindingly obvious


ivukotic (approved):
LGTM :)

PR #744, opened 2024-04-19
Larger batch

ivukotic:
this changes did-finder-lib used by rucio finder. it also updates version of rucio-client. This is needed in order to change batch size from 30 to 300.


BenGalewsky (approved):
Great!

PR #745, opened 2024-04-19
Increase rabbitMQ consumer timeout to one hour.

BenGalewsky:
During IDAP 200Gb challenge testing we found that very large datasets can take longer than 30 minutes to report from the DID finder back to the app. There is a default setting of 1/2 hour for a message to be ack-ed before closing the channel and putting the message back in the queue.

Increasing this value to 1 hour to avoid this problem.

### DIFF
--- a/helm/servicex/values.yaml
+++ b/helm/servicex/values.yaml
@@ -121,6 +121,9 @@ rabbitmq:
     enabled: false
   volumePermissions:
     enabled: true
+  extraConfiguration: |-
+    consumer_timeout = 3600000
+
 rbacEnabled: true
 secrets: null
 transformer:


ivukotic:
While this would prevent the issue we had last night (it took ~45 minutes), we did only one dataset that has 64k files. We will need to process 210k+ dataset files so just changing to 60 min won't really help. 
My PR that sends x10 more files per batch helps, but we should also make REST API faster. It is way too slow even with 5 pods each with 5 workers. We need to find what bottlenecks it.

BenGalewsky:
I'm going to merge this just so we can find this later if it comes up again

PR #748, opened 2024-04-21
Don't Allow Transformer Pod to Die While there are still files to upload to object stpore

BenGalewsky:
# Problem
As soon as a transform of a file is complete, the sidecar notifies the app. Meanwhile a separate thread is responsible for uploading the transformed files to Object Store. When the app decides that the last file has been transformed it sends the signal to shutdown all of the transformers. The problem is that the uploads to S3 may still be pending and so we lose files.

# Approach
The Kubernetes control plan sends a SIGTERM signal to the main process in the pod container to terminate the transformers. This PR adds a handler for this which sends a poison pill message through the queue to the object store uploader. When that thread reaches this item it shuts down cleanly.

Then transformer app now waits for this thread to exit before exiting which then clears the way for the pod to terminate.

We increase the termination grace period to 5 minutes to really make sure we can save all of the files. At this time we can also remove the sleep that was added between the last file processed notice being received and the transformers being shut down in the app.

In working on this PR I discovered that the RabbitMQManager constructor blocked. I made it it's own thread and start it too.

### DIFF
--- a/servicex_app/servicex/transformer_manager.py
+++ b/servicex_app/servicex/transformer_manager.py
@@ -26,20 +26,22 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import base64
-import os
-from typing import Optional
-import time
-
 import kubernetes
+import os
 from flask import current_app
 from kubernetes import client
 from kubernetes.client.rest import ApiException
+from typing import Optional
 
 from servicex.models import TransformRequest
 
 
 class TransformerManager:
     POSIX_VOLUME_MOUNT = "/posix_volume"
+
+    # This is the number of seconds tha thte transformer has to finish uploading
+    # objects to the object store before it is terminated.
+    POD_TERMINATION_GRACE_PERIOD = 5*60
 
     def __init__(self, manager_mode):
         if manager_mode == 'internal-kubernetes':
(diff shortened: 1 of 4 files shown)


ivukotic (commented):
I am not sure I understand how is this supposed to work. 
* will the grace period even be considered if there is no preStop hook defined?
* why do we need a poison pill if we are still deleting deployment?
it would be great if there is somewhere an explanation of this mechanism.
was this tested in real life?

BenGalewsky:
@ivukotic - some more context.

> * will the grace period even be considered if there is no preStop hook defined?

I learned quite a bit from the Kubernetes [pod termination lifecycle documentation](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination). The control plane sends the SIGTERM to the main process in the container and waits for that process to exit. If it doesn't exit before the grace period the control plane sends a KILL signal and immediately deletes the pod. So the grace period should be longer than we expect for the transformer to write the last files to object store, but we need to have it in case something goes wrong and the transformer gets into some illegal state from which it will never exit.

In the past we had no handler for that so the transformer exited immediately upon receiving the signal. No we receive the signal and begin an orderly shutdown so that the transformer script will only exit when it is safe to do so. We don't need the PreStop hook at all...

> * why do we need a poison pill if we are still deleting deployment?

The poison pill is sent to the thread inside the transformer pod that is uploading the files to object store. It instructs that thread to exit when it gets to that message. That's how we guarantee that the transformer script will only exit after the last file is uploaded. Perhaps when I said "queue" you were thinking of RabbitMQ... this is the thread safe queue we use for communication between the threads.

>   was this tested in real life?
I've tested this on my laptop and was able to confirm that the transformer pod doesn't get Killed by the Kubernetes control plane, it receives the SIGTERM and waits for the last file before exiting.

ivukotic:
while I am still not sure of ordering of poison pill sending and deployment deletion, if tested in real life I say let's merge it.

ivukotic (approved):
let's give it a try in production!

PR #757, opened 2024-04-30
Update the transform request document to show the title as well as a URL for the Logs

BenGalewsky:
# Problem
The client needs to know how to fetch error logs from Kabana. Also, when the client polls the transforms it doesn't get the transform title

# Approach
Add two fields to the transform request document


ivukotic (approved):
LGTM

PR #767, opened 2024-05-07
766 replica ordering

### LINKED ISSUE #766: replica ordering
Sometimes replica returned by Rucio are wrongly ordered. We need them resorted in transformer sidecar. 
The easiest way is to give a list of preferred endpoints as an environment variable. 
This would also help ServiceXLite.

ivukotic:
fixes issue with ?files etc..
added possibility to re-sort paths in transformers.
not sending full transformation log if it was successful.
ignoring errors on reading transformer log.

### DIFF
--- a/did_finder_rucio/scripts/did_finder.py
+++ b/did_finder_rucio/scripts/did_finder.py
@@ -63,6 +63,7 @@ def run_rucio_finder():
         logger.info('Starting rucio DID finder')
 
         async def callback(did_name, info):
+            did_name = did_name.split('?')[0]
             lookup_request = LookupRequest(
                 did=did_name,
                 rucio_adapter=rucio_adapter,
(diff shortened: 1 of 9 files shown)


BenGalewsky:
What was the issue with `?files` ? It looks like it's just been dropped entirely?

ivukotic:
> What was the issue with `?files` ? It looks like it's just been dropped entirely?

all the ?files and other modifiers were written in the DB table datasets as the unique dataset name. This made it so we had a lot of different entries and a lot of lookups for the same dataset (with 65k files). 
Did finder was anyhow doing the same job independently from these modifiers. This only affects what is written and read from the DB dataset table.

ivukotic:
now files=xxx works again

ivukotic:
It appears that there is not easy way of working around of the problem that wrong ?files=xxx implementation created.
So, I completely removed it. Now the dataset table in the database works correctly.

PR #770, opened 2024-05-14
Uproot-raw: copy histograms, import awkward functions for cuts

ponyisi:
Implements more sophisticated functions for cuts, allow copying of histograms for e.g. ATLAS TopCP output.

Details: 
* in the JSON query specifier for uproot-raw, a new `copy_histograms` key is added to specify that the specified object(s) should just be copied from the input to the output file. (If this is done with non-ROOT output file types, an exception will be thrown.)
* uproot interprets expressions passed to it for cuts using a sort of restricted Python, where in particular the available functions are restricted (by default to a small numpy subset). We add most of the useful awkward functions to the namespace, allowing the cut specifications to be much more powerful.

Like most codegens, it is not really possible to test this fully in the ServiceX repo only, since the generated code needs the physics image to execute. But we at least test the syntactic correctness of the generated code. I would claim that the tests should be done in the frontend library.


BenGalewsky:
It looks like this PR in introducing some keywords and special cases for the requests. Could you describe in the PR description what is changing and how this can be tested/used?

ponyisi:
@BenGalewsky , could you comment on whether the additional context in the description is enough? This PR is almost three weeks old, I would really like to be done with it.

BenGalewsky (approved):
Thanks for the additional context. I look forward to seeing this in action

PR #780, opened 2024-06-02
Log reading issue

ivukotic:
fixes it. removes sending full log for successful transforms


BenGalewsky (approved):
I pushed a commit removing the commented out code. Please don't check in commented out code

Issue #557, opened 2023-03-04
EOS Directory DID Finder


BenGalewsky:
There are some resources that are on EOS, but not registered with Rucio. This DID Finder will report all of the files in an EOS directory. A DID Path argument will specify whether to descend into subdirectories.

gordonwatts:
Let's see... I'm riffing here...

Here is a directory:
> `eos://eos/user/g/gwatts/public/data/mc16_13TeV.311423.MGPy8EG_A14NNPDF23_NNPDF31ME_HSS_LLP_mH600_mS150_lthigh.deriv.DAOD_EXOT15.e7357_e5984_s3234_r10201_r10210_p4696`

That directory should be visible publically, and has a single file in it.

And what if they want to run on one file?

> `eos:///eos/user/g/gwatts/public/data/mc16_13TeV.311423.MGPy8EG_A14NNPDF23_NNPDF31ME_HSS_LLP_mH600_mS150_lthigh.deriv.DAOD_EXOT15.e7357_e5984_s3234_r10201_r10210_p4696/DAOD_EXOT15.26710706._000001.pool.root.1`

Things to consider:

* Permissions - what is visible and what is not
* Is the URL above the right one? That is a CERN centric URL - should we be thinking about something more general (I do not know enough about EOS).
* How is EOS mounted from the DID container? Someone must know how to make EOS visible outside CERN, or is there a web API?

bbockelm:
Note "EOS" is the name of software while, conceptually, here you're only using the fact it's a filesystem remotely accessible over a protocol like WebDAV or XRootD.  There's several instance of this at CERN and maybe a dozen or two more outside CERN.

Might be useful to define scope here -- is it the concept of a remote filesystem?  Any XRootD endpoint?  Any EOS instance?  CERN EOS?  CERN ATLAS EOS?  Etc.

PR #790, opened 2024-06-12
Add XRootD DID finder

ponyisi:
Uses python xrootd library to expand a glob pattern given as an input. Addresses #557


BenGalewsky (approved):
This is so exciting! Let's deploy and see what we can learn

PR #793, opened 2024-06-16
Rename app's python package to servicex_app

BenGalewsky:
# Problem
At the time we migrated to the mono-repo, many of the services used the module name `servicex` - this wasn't an issue when the services were in their own repos, but this overlap really confuses IDEs and the code coverage tools.

# Approach
We've been slowly renaming packages. This PR is for the biggest one, the app. There are no changes to the functionality here, just renamed the package and updated the Docker file and the Gunicorn config file.


ponyisi:
Does there need to be a change of name inside the `[tool.coverage.run]` block inside the `servicex_app/pyproject.toml` as well?

ponyisi (commented):
See earlier question about the `pyproject.toml`, I'm happy to approve when answered.

ponyisi (approved):
Approved

PR #798, opened 2024-06-26
Try adding ARM64

ponyisi:
Try ARM64 image builds


ponyisi:
Any reason to not go ahead with this PR?

BenGalewsky:
Ok - looks like this doesn't add too much build time... let's see how it goes

BenGalewsky (approved):

Issue #807, opened 2024-07-12
UprootRaw needs protection against empty branch


kyungeonchoi:
**Describe the bug**
Transformation crashed if an accessed tree has no entry or no entry left after cut

**To Reproduce**
Deliver column (`el_pt`) from tree (`nominal`) from a file with no entry (`user.mgeyik:user.mgeyik.30182995._000042.out.root`)

**Expected behavior**
Transformation runs successfully and returns an empty array

**Additional context**
I haven't tested against Uproot FuncADL transformer yet but it might be the same situation

ponyisi:
A zero entry but well-defined tree should work already in the current version of the UprootRaw transformer. Please post the query that fails (preferably with a transformation ID).

PR #810, opened 2024-07-15
Fix for #807

ponyisi:
When using `uproot.iterate` with an empty tree, or one where the cuts remove all output, the existing code will encounter a condition where it tries to take the length of `None`. Protect against this, and also update later logic since newer uproot versions will handle `arrays` on an empty tree properly.

Also introduce workarounds for `uproot` 5.3.9 handling of colons in filenames (it incorrectly applies URL cleanups).


BenGalewsky (approved):
Thanks -- it would be super cool if we had unit tests for the templated code

kyungeonchoi (approved):

PR #811, opened 2024-07-16
Fully new cleanup service

ivukotic:
This cleans both S3 and the database.


cleanup/scripts/cleanup.py
```diff
@@ -212,32 +171,25 @@ def run_minio_cleaner():
...
         sys.exit(1)
 
-    try:
-        if 'MINIO_ENCRYPT' in os.environ:
-            if isinstance(os.environ['MINIO_ENCRYPT'], bool):
-                use_https = os.environ['MINIO_ENCRYPT']
-            else:
-                use_https = strtobool(os.environ['MINIO_ENCRYPT'])
+    if 'S3_ENCRYPT' in os.environ:
+        if isinstance(os.environ['S3_ENCRYPT'], bool):
```
ponyisi, line 175: can this ever be bool? shouldn't os.environ only ever have strings?

cleanup/servicex_storage/db_manager.py
```diff
@@ -0,0 +1,159 @@
...
+            req_age = datetime.datetime.now(
+                datetime.timezone.utc) - req.submit_time.replace(tzinfo=datetime.timezone.utc)
+            if req_age.days > max_request_age:
+                self.delete_request(req)
+
+        # get all the datasets in ServiceX
+        datasets = self.get_datasets()
+        self.logger.info('datasets in servicex db', extra={"datasets": len(datasets)})
+
+        # delete datasets not used recently
```
ponyisi, line 152: Is there a possible race condition here? (between lookup of datasets and deletion, a job starts and requests the dataset that is about to be cleaned?)

BenGalewsky:
We can close this now since we've merged #1166

PR #819, opened 2024-08-02
Make celery worker dockerfiles work on OpenShift

BenGalewsky:
# Problem
OpenShift creates unique users to run inside the pods. This disables all user management configured in the Dockerfile. DID Finder pods were all failing with permission errors.

# Approach
Following this [OpenShift guide](https://developers.redhat.com/blog/2020/10/26/adapting-docker-and-kubernetes-containers-to-run-on-red-hat-openshift-container-platform#) we use the assumption that no matter who the user is, OpenShift will run the container in group 0.

Updated the Dockerfiles:
1. Create the celery user  in group 0 to insure compatibility with non-openshift clusters
2. Change the group ownership of the /opt/servicex directory to zero 
3. Update read and write permissions to include anyone in group 0
4. Make sure that all poetry file and directory creation occurs as root, before we change to the `celery` user and update permissions (some files were being created after this and didn't get updated permissions)
5. During testing, I became suspicious of putting the source code in `/usr/app` and moved everything to the more expected `/opt/servicex` directory.

### DIFF
--- a/did_finder_cernopendata/Dockerfile
+++ b/did_finder_cernopendata/Dockerfile
@@ -5,10 +5,10 @@ LABEL maintainer="Gordon Watts <gwatts@uw.edu>"
 USER root
 
 # Create app directory
-WORKDIR /usr/src/app
+WORKDIR /opt/servicex
 
-# Create celery user
-RUN useradd -ms /bin/bash celery
+# Create celery user. Assign them to group zero since that is the group OpenShift will run the container as
+RUN useradd -g 0 -ms /bin/bash celery
 
 ENV  POETRY_VERSION=1.5.1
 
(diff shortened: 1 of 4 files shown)


did_finder_cernopendata/Dockerfile
```diff
@@ -25,19 +25,23 @@ COPY poetry.lock poetry.lock
...
-RUN chown -R celery:celery /usr/src/app
+
+ENV XDG_CONFIG_HOME=/opt/servicex
+RUN poetry config virtualenvs.in-project true
+RUN poetry install --no-root --no-interaction --no-ansi
+
+# Change ownership of the app directory to celery user. Also set the group to zero since
+# that is the group OpenShift will run the container as
+RUN chown -R celery:0 /opt/servicex
+RUN chgrp -R 0 /opt/servicex && chmod -R g=u /opt/servicex
```
holzman: I think the  `chgrp -R 0` is redundant since the `chown` call above specifies the group.  (I only noted this once but it's the same in all three).

holzman (commented):
LGTM, modulo the note about redundant code.  I think moving stuff to /opt/servicex is a good idea as well.

holzman:
Just tested: `did_finder_rucio/scripts/start_celery_worker.sh` needs to be updated for the new location of `proxy-exporter.sh`.

did_finder_cernopendata/Dockerfile
BenGalewsky: Tried removing it along with the `g=u` chmod and it still works! Delete the entire line

BenGalewsky:
> Just tested: `did_finder_rucio/scripts/start_celery_worker.sh` needs to be updated for the new location of `proxy-exporter.sh`.

Oops - fixed that

PR #826, opened 2024-08-21
Use Celery to Submit Transforms to execute in sidecar

BenGalewsky:
# Problem
The transformer sidecar uses a home-made RabbitMQ interface. This makes it harder to maintain and also uses a callback mechanism that doesn't lend itself to testing. Furthermore, the service is hard to debug and has a transaction hole in between the time the sidecar reports the transform complete and before it is uploaded to minio

# Approach
Migrate the service to be a celery worker. This becomes easier to write unit tests since the Celery task, `transform_file` is also just a python function that can be called in unit tests.

Took advantage of this opportunity to update and correct a number of other issues with the transformer:
1. Migrate all of the science container socket-based communication to its own class to clean up the sidecar code and make it more testable.
2. Move reporting of successful transforms into the object store uploader (unless we are just writing to a posix volume)
3. Setting auto-delete on rabbit queues associated with the transformer
4. Cleaned up the command line arguments to make their intention clearer
5. Removed code to allow the sidecar to be used from the command line to transform a single file. This is better handled by creative use of the science container
6. The watch script accepts a new argument which is the host name to bind the communications port to the sidecar on. This allows for users to run the science container in a docker image, and connect to a debuggable sidecar running on the users laptop.
7. Cleaned up some logging chatter
8. Added more support for testing out the sidecar in your local IDE
9. Handle an additional uproot error case


ponyisi:
Can we merge this branch? It seems quite urgent for client stability.

helm/scripts/port-forward.sh
```diff
@@ -18,7 +18,7 @@ stop_port_forward() {
 # Start port forwarding for all services
 start_all() {
   start_port_forward "servicex-ben-postgresql" "5432:5432"
-  start_port_forward "servicex-ben-code-gen-uproot" "5001:8000"
```
ponyisi, line 21: I'm a bit confused about this port forward - is it needed for the func_adl uproot codegen (in which case we should keep the old line too)? In any case the uproot-raw codegen doesn't expose anything on port 5001?

ponyisi:
@BenGalewsky - could you document somewhere (this PR, a README, I dunno) the communication path in play now? (we start a transformation, a celery queue is set up, the sidecars subscribe ...) just so we have this on the record. Given that and the answer to the one query above I will approve, since this at a minimum makes our tests work where `develop` currently does not.

helm/scripts/port-forward.sh
BenGalewsky: This little script is what I use during development. All it does is port forward the pods running on my laptop's Kubernetes cluster so I can run the app in my IDE from outside the cluster. I only have room to run one code gen in this little deployment, so I just choose one.

BenGalewsky:
> @BenGalewsky - could you document somewhere (this PR, a README, I dunno) the communication path in play now? (we start a transformation, a celery queue is set up, the sidecars subscribe ...) just so we have this on the record. Given that and the answer to the one query above I will approve, since this at a minimum makes our tests work where `develop` currently does not.

Added info in the architecture docs (really the whole doc is out of date, but the rest of it will have to be fixed another time...)

ponyisi (commented):
Approved

PR #835, opened 2024-09-09
[WIP] Documentation pass

LincolnBryant:
This PR is tracking documentations as I go through the docs as a first time user. I will also put some commentary here as I go.

1. Updated the suggested Helm chart version to v1.4.1 in the deployment guide* 

*One thing that would be useful to me is to know what versions are available without installing the Helm tools. I snooped around and found the list of releases here: https://github.com/ssl-hep/ssl-helm-charts/tree/gh-pages

### DIFF
--- a/docs/deployment/basic.md
+++ b/docs/deployment/basic.md
@@ -136,7 +136,7 @@ x509Secrets:
 Deploy the chart to your cluster with
 
 ```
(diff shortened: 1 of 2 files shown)


BenGalewsky:
Superseded by @MattShirley 's PR

PR #842, opened 2024-09-14
Stop celery eating all the sidecar logging

ponyisi:
This fixes an issue where celery changes the root logger for tasks, meaning that the sidecar output was being swallowed. Separately a small fix for more information in transformation submission.


BenGalewsky:
We can close this without merging, right @ponyisi ?

ponyisi:
> We can close this without merging, right @ponyisi ?

Yes, should be redundant now.

Issue #843, opened 2024-09-15
Celery-enabled transformers lose files due to queue auto-deletion


ponyisi:
**Describe the bug**
When run at large scale, transformation requests are lost because RabbitMQ queues are auto-deleted before all files are complete. This appears to be because the auto-delete behavior depends on whether there are consumers connected, _not_ whether the queue is empty, and so if the workers disconnect while they are working on a long-lived task, RabbitMQ can decide that nobody is interested in the queues any more and delete them.

**To Reproduce**
1. Run the ServiceX client 3.0 version of the CMS AGC grand challenge (https://github.com/iris-hep/analysis-grand-challenge/pull/225) with no restriction on the number of files (i.e. the number of files to transform exceeds the number of ServiceX transformation workers). The jobs will hang part way through. The disappearance of the relevant queues while the jobs are still ongoing is verified using `rabbitmqctl list_queues` in the RabbitMQ pod, and when the auto_delete option is removed, the transformations fully complete as expected.

**Expected behavior**
Transformations complete (equivalently, the transformer queues are only deleted at the end of the transformation).

PR #844, opened 2024-09-16
Fixes for celery

ponyisi:
A number of fixes for the new celery-based transformer sidecars.
* Incorporate #842 
* Fix a bug in the sidecar-science image communication (a missing period) which was leading to the stop signal being treated as an (invalid) transform request
* Do not use `celery_app.control.broadcast('shutdown')` : this terminates *all* Celery workers, including those in other transforms and the DID finders (explaining many weird glitches with submissions and the tendency for the `auto_delete` queues to vanish when multiple transforms were in flight [#843]). Instead send a termination message only to the workers of a specific queue, when the transformation is complete, using `celery_app.control.cancel_consumer`

This is tested extensively on -testing3 using the CMS AGC notebook, where it seems quite rock-solid (as opposed to devel without these fixes, which basically never completes).

### DIFF
--- a/servicex_app/servicex_app/celery_task_router.py
+++ b/servicex_app/servicex_app/celery_task_router.py
@@ -36,8 +36,7 @@ def route_task(name, args, kwargs, options, task=None, **kw):
         return {
             "queue": Queue(name=f"transformer-{kwargs['request_id']}",
                            durable=False,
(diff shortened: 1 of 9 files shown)


BenGalewsky (approved):
Wow, thank you @ponyisi

PR #845, opened 2024-09-16
Improve robustness of uproot-raw code with retries

ponyisi:
XrootD "Operation expired" errors can occur on `uproot.open`, and are transient. These errors occur at e.g. a few percent level when directly reading files from CERN EOS from the SSL, and if a single file fails the entire request is generally useless. Using a retry mechanism allows us to make the system more robust in this mode. (There may be distinct types of problem when using xcache, this needs to be studied.)

This PR is tested using the CMS AGC notebook on a testing instance with xcache switched off and files from CERN EOSPUBLIC. Tenacity is included via the science image.


BenGalewsky (approved):
This makes sense - if we can retry down here it is more efficient than retrying from the sidecar

ponyisi:
> This makes sense - if we can retry down here it is more efficient than retrying from the sidecar 

Yeah, the main advantage here is that we can trap very specific errors that we can expect to be transient, which would be hard to reliably detect above.

PR #847, opened 2024-09-19
Inclusive error logging

ponyisi:
Restore adding the full log body from a failed transformer to what gets logged in Kibana. Also improves the uproot-raw error reporting, very slightly (shows the exception message, not the line that raised the exception).


BenGalewsky (approved):
Thanks

Issue #724, opened 2024-04-08
The "Status" on the dashboard does not update when moving out of the lookup state


gordonwatts:
**Describe the bug**
If you are looking at the SX Dashboard, and looking at the status column, it will not update out of the `lookup` state.

**To Reproduce**
1. Go to your dashboard.
2. Start a SX query - on a new dataset (otherwise it is too fast!!)
3. Reload your dashboard page before the lookup is completed.
4. Note that even after files are started - or the transform finishes - that status does not update.

**Expected behavior**
Expect the status to track the actual status.

N.B. When the transform finished the status does update to _Completed_.

**Screenshots**
![image](https://github.com/ssl-hep/ServiceX/assets/1778366/5e1a794f-b4fc-447a-b2ec-7cd325c5d78f)

PR #849, opened 2024-09-21
Fix the Eternal Lookup of the Rucio Dataset

ponyisi:
Do the status change from lookup to running when the dataset is complete


BenGalewsky (approved):
Thank you for finding this!
In future it would be helpful to have a little more description in the PR description to explain what you are fixing and how you fixed it

ponyisi:
This resolves #724

PR #853, opened 2024-09-24
Update the chart's values.yaml to use released images

### LINKED ISSUE #851: Update CI action to publish helm chart to servicex-prod instead of -int
We would like to have the `servicex-prod` SSL instance be the one that holds the latest production release, and `servicex-int` the one that runs the `develop` version. We should therefore update https://github.com/ssl-hep/ServiceX/blob/develop/.github/workflows/ci_helm_publish.yaml to publish to `servicex-prod` instead of `servicex-int`.

### LINKED ISSUE #852: ci_helm_publish Needs to Update Image Tags in Values.yaml
# Problem
When we publish a new helm chart, the bundled values.yaml has `develop` tags for all images. This needs to be updated to point to the new version's images

BenGalewsky:
# Problems
1. When a helm chart is released, the docker image tags in `values.yaml` are stuck pointing to `develop` images instead of the tagged images for the release.
2. Releases of the service are deployed to _ServiceX-int_ namespace (which is broken) instead of to `-Prod` namespace.

Resolves #852 
Resolves #851 

# Approach
For (2) simplified the values update step and directed it to the -prod values

For (1) I dug deeper into the yq expressions and performed the transformations on the checked out `values.yaml`

### DIFF
--- a/.github/workflows/ci_helm_publish.yaml
+++ b/.github/workflows/ci_helm_publish.yaml
@@ -40,18 +40,34 @@ jobs:
           echo "DOCKER_TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
           echo "$DOCKER_TAG"
 
-      - name: Set version in flux_river_configs
+      - name: Set version in helm values.yaml
+        # Update the tags for images in the helm Chart.yaml and values.yaml to match
+        # The deployed docker image tag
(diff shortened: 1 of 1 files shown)


ponyisi (approved):
Looks good to me. `yq` is very cool!

PR #855, opened 2024-09-25
Fix pending lookup issue #854

### LINKED ISSUE #854: Celery DID finder has an invalid call on fileset_complete
`fileset_complete.py` has the following code:

```
            for transform_request in TransformRequest.lookup_pending_on_dataset(int(dataset_id)):
                dataset.publish_files(transform_request, self.lookup_result_processor)
                transform_request.status = TransformStatus.running
```

However `dataset` does not have the `publish_files` method, this is actually a method on `DatasetManager` and not `Dataset`. Whenever this codepath is hit, if there are pending lookups, the pending lookups will not be updated (and neither will the primary request!) (Not sure how this passed the linting ...)

ponyisi:
Fix #854 : the method `publish_files` was being called on the wrong object.

### DIFF
--- a/servicex_app/servicex_app/resources/internal/fileset_complete.py
+++ b/servicex_app/servicex_app/resources/internal/fileset_complete.py
@@ -28,6 +28,7 @@
 from flask import request, current_app
 
 from servicex_app.models import Dataset, db, TransformRequest, TransformStatus, DatasetStatus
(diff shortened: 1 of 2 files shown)


BenGalewsky (approved):
Thanks for catching this

PR #856, opened 2024-09-25
Use Rucio's site-closeness algorithm for replica ordering

ponyisi:
Propagate the environment variable SITE_NAME into the Rucio DID finder container, if `didFinder.rucio.site` is set in the `values.yaml`. If available, provide this to the rucio client to prioritize replicas.


BenGalewsky (approved):
How to improve rucio performance with this one cool trick!

PR #860, opened 2024-09-26
Migrate all of the codegens from -prod into values.yaml

BenGalewsky:
The chart's values.yaml will serve as the definitive list of codegens and we can keep -prod values.yaml sweet and simple


ponyisi (approved):
Looks good

PR #864, opened 2024-09-27
Apply timeouts and retries to all usage of requests

ponyisi:
It is highly recommended to apply a timeout when using `requests`. This will e.g. allow it to fall back to IPv4 if IPv6 is broken, or to use a second server in a high-availability cluster if the first one returned by DNS is not accepting requests. Also add retry protection to avoid immediately dying if a timeout is encountered, and a metric for how long uploads to the object store take.

Only a connection timeout is implemented, so we do not accidentally retry requests just because the called services are a bit slow. Tested with my usual testsuite on testing-3.

### DIFF
--- a/servicex_app/servicex_app/reliable_requests.py
+++ b/servicex_app/servicex_app/reliable_requests.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2019, IRIS-HEP
+# Copyright (c) 2024, IRIS-HEP
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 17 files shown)


gordonwatts:
There seems to be a fair amount of cut paste code to get the retry in there. Should that be something more generic?

Also, does this solve the ipv6 vs ipv4 problem? The retry has no logic to switch interfaces, does it?

Finally, it does feel like we are adding code to work around a badly configured site. Is that worth it? Better to get the site to fix it. Partly I say this because there are likely an infinite number of ways to badly configure a site.

BenGalewsky:
This seems like a good practice.. I suppose we could have our own wrapper to `post` .. also it's so much cleaner to isolate the requests into a function and use the `@retry` decorator.

Finally, it's not nice to add unrelated changes to a PR (I'm looking at you, elapsed time logging)

ponyisi:
@gordonwatts in fact it seems `requests`/`urllib3` does try IPv4 if IPv6 fails, at least in these cases - but the connection needs to time out first. So our retry doesn't do anything to help with the IPv6 problem, but on connection timeout the underlying library will first try another connection on a different IP before reporting an error that would be retried.

@BenGalewsky happy to remove the elapsed time logging (it is a way of tracking whether or not the timeouts are working in a bad environment though, which is why I left it in)

servicex_app/servicex_app/reliable_requests.py
```diff
@@ -0,0 +1,54 @@
...
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+from functools import wraps
+
+from tenacity import retry, stop_after_attempt, wait_exponential_jitter
+
+# This is the default timeout settings for requests. It represents the time to make a
+# connection and then the time to wait for a response.
+request_timeout = (0.5, None)
```
ponyisi: Standard Python style would make this `REQUEST_TIMEOUT` in caps?

BenGalewsky: Ah yes.. done

BenGalewsky (approved):
Hooray for reliability

PR #867, opened 2024-09-27
Update values.yaml to address #866

### LINKED ISSUE #866: Indentation of v1.5.1 helm chart is wrong for some codegens
Several codegens are disabled due to over-indentation in the default values.yaml file in v1.5.1 (therefore are not deployed by default). These are `atlasr21`, `atlasr22`, and `cmssw-5-3-32`.

ponyisi:
Fix indentation of certain codegens in values.yaml (resolves #866)


BenGalewsky (approved):
Thanks - this is what happens when I attempt to create a new release without testing it as a pre-release deployment!

PR #881, opened 2024-10-11
transform_file task uses ack_late=True

BenGalewsky:
We found that when the autoscaler decides to scale down the number of transformers that the science image dies immediatly which  causes the "sync" call to fail. When this happens we need to put the transform request back onto rabbit queue and die gracefully.

Add the acks_late propoerty to the transform_file request and now the failure to receive the sync command from the science container is a fatal error.

### DIFF
--- a/transformer_sidecar/src/transformer_sidecar/object_store_uploader.py
+++ b/transformer_sidecar/src/transformer_sidecar/object_store_uploader.py
@@ -27,6 +27,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import logging
 import os
+import signal
 from multiprocessing import Process
 from pathlib import Path
 from queue import Queue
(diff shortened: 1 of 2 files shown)


transformer_sidecar/src/transformer_sidecar/object_store_uploader.py
```diff
@@ -58,12 +59,19 @@ class ObjectStoreUploader(Process):
...
         self.request_id = request_id
         self.input_queue = input_queue
         self.logger = logger
         self.convert_root_to_parquet = convert_root_to_parquet
 
+        signal.signal(signal.SIGTERM, self.handle_sigterm)
+
+    def handle_sigterm(self, signum, frame):
+        # This method will be called when SIGTERM is received
+        print("SIGTERM received, but ignored.")
```
ponyisi: Can this log instead of print? Even better if we can provide a list of the objects remaining to be uploaded as an extra.

BenGalewsky: Do all `extra`s make it into the log? I've felt like there is some sort of schema that keeps arbitrary info out

ponyisi: I've always seen the extras that I'm looking for...

BenGalewsky: Fixed

PR #885, opened 2024-10-14
Revert to always using internal cluster interface between transformer and app

### LINKED ISSUE #884: Don't use ingress route for internal communications
# Problem
The `ADVERTISED_HOSTNAME` setting in the app config is the name the transformers use to communicate back to the App. As part of ServiceXLite, this is the external name of the ingress. According to a [web search](https://www.perplexity.ai/search/in-kubernetes-if-a-pod-wants-t-gQKujOreQ3K9pFVEkQkg7A), this causes the traffic to be routed through the internet instead of the internal Kubernetes service.

# Approach
Until we rewire the internal communications to use RabbitMQ, we need to revert to the transformers using the internal name to communicate with the app

BenGalewsky:
Resolves #884

### DIFF
--- a/helm/servicex/templates/app/configmap.yaml
+++ b/helm/servicex/templates/app/configmap.yaml
@@ -65,13 +65,8 @@ data:
     RABBIT_RETRIES = {{ .Values.app.rabbitmq.retries }}
     RABBIT_RETRY_INTERVAL = {{ .Values.app.rabbitmq.retry_interval }}
 
-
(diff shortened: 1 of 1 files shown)


ponyisi:
So I think the claim here would be that any federated instances would have to be given an explicit override for the endpoint URL, which I think is fine. And eventually I guess these callbacks will all migrate to Celery anyway.

ponyisi (approved):

PR #891, opened 2024-10-16
Point to develop for all codegens so releases build properly

ponyisi:
Don't reference the 1.5.1-alpha.1 tag in the `values.yaml`; we need `develop` in order for the tags to be updated correctly when we build the final helm chart for a release.


BenGalewsky:
Should I just build a 1.5.2-post.1 helm chart so we don't have to resort to vandalizing develop?

BenGalewsky (approved):
Ah, I get this now... I was reading it backwards and thought you were pointing to the versioned images instead of reverting them to `develop`

Issue #889, opened 2024-10-16
Add `values.yaml` option to pass database connection configuration arguments to SQLAlchemy


ponyisi:
Story: As a site administrator or developer, I would like to test different configuration options for the Postgres library when connecting to the backend database without having to build new images hardcoding in those options.

How to implement: in the `values.yaml` allow the specification of the `SQLALCHEMY_ENGINE_OPTIONS` argument for Flask. These will be passed as `engine_options` to SQLAlchemy.

BenGalewsky:
Closed via #892

PR #892, opened 2024-10-18
Permit passing SQLAlchemy configuration options

ponyisi:
Allow sites to specify configuration for SQLAlchemy connections as a Python dict; for example, this can be used to experiment with the preping option for checking connections.


ponyisi:
Resolves #889

BenGalewsky (approved):

PR #896, opened 2024-10-22
Rucio replica sorting service using geolocation

ponyisi:
This uses functionality added to the DID finder library (https://github.com/ssl-hep/ServiceX_DID_Finder_lib/pull/34) to sort replicas for the Rucio DID finder. In principle a similar pattern can be implemented for other DID finders, though none of the others provide multiple replicas at the moment.


BenGalewsky:
@ponyisi will you replace this PR with a simplified one that is just implemented in the Rucio DID finder?

ponyisi:
I will "un-draft" this once the DID finder library is updated and I don't have to point at a git branch any more... but I've verified it works OK.

ponyisi:
Resolves #766

BenGalewsky:
@ponyisi -- looked at the code in the DID Finder Library - Given the [YAGNI](https://www.geeksforgeeks.org/what-is-yagni-principle-you-arent-gonna-need-it/) principle  I would prefer that we implement this in the Rucio DID finder to start with. We are going to find problems and enhancements and fixing things in the library is a drag.

I would prefer to see this initial implementation entirely done in the Rucio DID Finder and then we can move it to the library once we find a use case for that.

ponyisi:
@BenGalewsky whatever's needed to get this in ... it's rather desperately needed....

did_finder_rucio/src/rucio_did_finder/replica_distance.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        if db_url_tuple is None:
+            db_url_tuple = self.get_download_url_from_environment()
+        self._download_data(db_url_tuple)
+
+    def sort_replicas(self, replicas: List[str], location: Mapping[str, float]) -> List[str]:
+        """
+        Main method of this class.
+        replicas: list of strings which are the URLs for the replicas for a file
+        location: dict of the form {'latitude': xxx, 'longitude': yyy} where xxx and yyy are floats
```
BenGalewsky, line 96: Any reason lat and long can't be instance variables?

did_finder_rucio/src/rucio_did_finder/replica_distance.py
```diff
@@ -0,0 +1,159 @@
...
+
+@lru_cache
+def _get_distance(reader: Optional[geoip2.database.Reader],
+                  fqdn: str, my_lat: float, my_lon: float):
+    """
+    Determine angular distance between server at fqdn and (my_lat, my_lon).
+    If there is a failure of fdqn location lookup, will return pi
+    """
+    if reader is None:
+        return math.pi
```
BenGalewsky, line 65: Why pi?

did_finder_rucio/src/rucio_did_finder/replica_distance.py
```diff
@@ -0,0 +1,159 @@
...
+    dellat = math.radians(lat2-lat1)
+    dellon = math.radians(lon2-lon1)
+    hav_theta = ((1-math.cos(dellat))/2 +
+                 math.cos(math.radians(lat1))*math.cos(math.radians(lat2))*(1-math.cos(dellon))/2)
+
+    return 2*math.asin(math.sqrt(hav_theta))
+
+
+@lru_cache
+def _get_distance(reader: Optional[geoip2.database.Reader],
```
BenGalewsky: Reader is such a confusing name for this... it's a populated geoip_database

did_finder_rucio/src/rucio_did_finder/replica_distance.py
```diff
@@ -0,0 +1,159 @@
...
+        replicas: list of strings which are the URLs for the replicas for a file
+        location: dict of the form {'latitude': xxx, 'longitude': yyy} where xxx and yyy are floats
+        giving the latitude and longitude in signed degrees
+        """
+        if not self._reader:
+            return replicas
+        if len(replicas) == 1:
+            return replicas
+        fqdns = [(urlparse(replica).hostname, replica) for replica in replicas]
+        distances = [(_get_distance(self._reader, fqdn, location['latitude'],
```
BenGalewsky: This statement could be a lot simpler if we had a class to hold each FQDN, and `distance` could be a method on it. Then we could just use a sorted list

did_finder_rucio/src/rucio_did_finder/replica_distance.py
ponyisi: Because pi is the largest possible result of the haversine computation (default is "the farthest possible location")

did_finder_rucio/src/rucio_did_finder/replica_distance.py
ponyisi: Fine, we can change it

did_finder_rucio/src/rucio_did_finder/replica_distance.py
ponyisi: When originally written, the idea was you could have one ReplicaSorter which would be a service accepting distance sorting requests from clients (i.e. subsequent calls could have different locations), so this is a bit of an artifact. That said since this is now internal to the Rucio DID finder I don't think it's particularly interesting where exactly we store the numbers.

did_finder_rucio/src/rucio_did_finder/replica_distance.py
ponyisi: disagree that it would be technically simpler. The "class" would just be the moral equivalent of a `namedtuple` and we would have to instruct the sorting of what key to sort on. I suppose it could make it really explicit what is happening here, though what we have here is a pretty common Python idiom.

BenGalewsky (approved):
Thanks for the changes. I think we have different ideas about code clarity, but this should be ok to maintain

PR #897, opened 2024-10-22
Limit the autoscaling of transformations to the number of files in a request

ponyisi:
Since we have file granularity for transformations, it does not make sense to autoscale beyond the total number of files in a transform request. In fact this can be damaging since the autoscaler may be too enthusiastic at first, then scale the pod down, killing running jobs. Provide a hint to not scale too aggressively.

* datasets with known sizes have the max number of replicas set immediately

No action is taken for datasets which have to be looked up, since we don't know how many files there will be, and we do not want to shut down pods by force if we find we've spawned too many once the lookup is done (let the autoscaler take the blame for this one).

### DIFF
--- a/servicex_app/servicex_app/resources/servicex_resource.py
+++ b/servicex_app/servicex_app/resources/servicex_resource.py
@@ -31,7 +31,7 @@ import pkg_resources
 from flask import current_app
 from flask_jwt_extended import get_jwt_identity, jwt_required
 from flask_restful import Resource
-from servicex_app.models import UserModel, TransformRequest
+from servicex_app.models import UserModel, TransformRequest, TransformStatus
 
 from servicex_app.transformer_manager import TransformerManager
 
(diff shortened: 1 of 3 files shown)


servicex_app/servicex_app/resources/servicex_resource.py
```diff
@@ -92,6 +93,9 @@ def start_transformers(cls, transformer_manager: TransformerManager,
         transformer_manager.launch_transformer_jobs(
             image=request_rec.image, request_id=request_rec.request_id,
             workers=request_rec.workers,
+            max_workers=(max(1, request_rec.files)
+                         if request_rec.status == TransformStatus.running
```
BenGalewsky, line 97: Shouldn't this be limited to `TRANSFORMER_MAX_REPLICAS` in case the number of files is > than that number?

ponyisi: Limit is applied in the autoscaler invocation https://github.com/ssl-hep/ServiceX/pull/897/files#diff-ab4dc02ce708be56b4ebde72961b66d0fabf04bc255fdb60154f88889711edc5R333

ponyisi: I _think_ this point is addressed? (the max limit is applied during the HPA definition.) Let me know if not.

BenGalewsky (approved):
Thanks for finding your way through this...

PR #898, opened 2024-10-24
Switch default atlasr21 image to something actually versioned

ponyisi:
Change the image reference `sslhep/servicex_code_gen_func_adl_xaod` to `sslhep/servicex_code_gen_atlas_xaod` (which is what we are calling it now, and which is built as part of a release) for `atlasr21` in the values.yaml


BenGalewsky (approved):
Thanks

PR #907, opened 2024-11-06
Add route to list datasets

BenGalewsky:
As an initial step to being able to manage cached datasets, we need an ability to report on the datasets in the system.

Adds a REST endpoint that returns datasets as a JSON list. The endpoint accepts an optional argument to filter on DID Finder (i.e. 'rucio' or 'user')


ponyisi (approved):
Looks good

PR #909, opened 2024-11-12
Give tag of XRootD DID finder image in about page

ponyisi:
Update about page template to give XRootD DID finder image tag


BenGalewsky (approved):
Yes!

PR #910, opened 2024-11-12
Reduce number of logging messages from Celery operations

ponyisi:
Celery by default is quite noisy (e.g. starting and stopping queues results in lots of broadcast messages). Try to reduce this a bit.


BenGalewsky (approved):
Shush!

PR #912, opened 2024-11-13
More concurrency for Rucio DID lookups

ponyisi:
Increase concurrency for Rucio lookups to allow five in parallel (allow things to proceed even if there is a long-running job)


BenGalewsky (approved):
Let's see what this does for us

PR #913, opened 2024-11-13
Raise ack timeout for transformers to one month

### LINKED ISSUE #893: Files are lost when RabbitMQ timeouts are hit
When we have very long transforms, they can exceed the timeout set by RabbitMQ for sending ACK messages (3600000 ms = 1 hour). This has a couple of consequences:
* the transform never gets acknowledged at all (errors like
```
Unrecoverable error: PreconditionFailed(406, 'PRECONDITION_FAILED - delivery acknowledgement on channel 1 timed out. Timeout value used: 3600000 ms. This timeout value can be configured, see consumers doc guide to learn more', (0, 0), '')
```
* Even in the best case scenario (Celery retries the task because the lack of ACK pushes it back on the queue) the transformation can never complete, because it will generally always timeout;
* I'm not sure that Celery actually does this anyway - I think RabbitMQ drops the connections to the transformer pods when the timeouts happen, and it seems that we can wind up in a situation where all the connections are closed simultaneously, causing the queue to be deleted (even if it still has tasks in it).

I would suggest either disabling timeout completely on transformer job queues, or setting some really long (site-configurable) timeout and explicitly mentioning that overrunning tasks may be canceled without warning.

ponyisi:
Resolves #893: changes queue configuration for transformer queues so that ACK timeout intervals are now the very long length of one month.

### DIFF
--- a/servicex_app/servicex_app/celery_task_router.py
+++ b/servicex_app/servicex_app/celery_task_router.py
@@ -36,7 +36,8 @@ def route_task(name, args, kwargs, options, task=None, **kw):
         return {
             "queue": Queue(name=f"transformer-{kwargs['request_id']}",
                            durable=False,
-                           auto_delete=True)
+                           auto_delete=True,
+                           queue_arguments={'x-consumer-timeout': 2_629_746_000})
         }
 
     did_finder_match = re.search(did_finder_pattern, name)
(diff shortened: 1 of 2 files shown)


BenGalewsky (approved):
See you in a month

PR #914, opened 2024-11-13
Mark Datasets as Stale to Force Requery

### LINKED ISSUE #850: Refresh Cached Datasets
_As a ServiceX analyzer I want to be able to refresh the list of replicas in a cached dataset so I can work around invalid replicas_

# Description
Caching datasets has improved performance of the system since we no longer look up DIDs once they have been resolved once. There are cases where the list of replicas in the cache are no longer valid and we need a way to refresh them.

# Approach
From a data integrity point of view it makes sense to keep the old replica records around since they were used in generating existing transforms. We have foreign key constraints on the `files` and `transform_result` tables, so simply deleting the dataset will cause foreign key violations without deleting the old transforms and their results as well. These constraints help keep the database consistent.

## Modify the Datasets Table
We can add a new column in the datasets table to record if a did is stale and needs to be looked up again the 
next time it is requested. 

1.  Add a new boolean value to the `datasets` 
```
is_stale = db.Column(db.Boolean, default=False)
```

2. Update the `find_by_name` method to only return rows where `is_stale` is False

## Add New Endpoints
We will add new REST endpoints to allow users to manipulate the stale setting. For usability we should also have an endpoint that lists datasets.

1. GET on `/dataset` to return a listing of all datasets along with the some statistics 
2. DELETE on `/dataset/{id}` to mark the dataset as stale and force a re-query next time

## New CLI Commands
Add new commands to the CLI to allow users to easily interact with these new endpoints

```
servicex dataset list
servicex dataset delete
```

BenGalewsky:
# Problem
Once a dataset is cached in the service there is no way to evict them from the cache and force a reload. This can cause problems if the replicas move or if the dataset is updated.

Resolves #850 

# Approach
1. Add a new value to the dataset table, `stale` - this is set to True to evict a dataset from the cache
2. The find_by_name and get_all methods on Dataset model now filter out datasets where stale is True
3. Updated the data model so that did_name is no longer unique
4. Added REST endpoints to delete a dataset (which sets stale to True)
5. During development it made sense to add an endpoint to get the details of a single dataset
6. Added an argument to the REST call that retrieves datasets to also show the deleted datasets with stale=True

### DIFF
--- a/servicex_app/migrations/versions/v1_5_5.py
+++ b/servicex_app/migrations/versions/v1_5_5.py
@@ -21,10 +21,16 @@ def upgrade():
                                         sa.Boolean(),
                                         nullable=False,
                                         server_default='false'))
+
+    # Name is no longer unique, so we need to drop the unique index
+    # op.drop_index('datasets_name_key', table_name='datasets')
+
+    op.drop_constraint('datasets_name_key', 'datasets')
     # This was never used, so a fine time to delete it
     op.drop_column('transform_result', 'did')
 
 
 def downgrade():
+    op.create_unique_constraint('datasets_name_key', 'datasets', ['name'])
     op.drop_column('datasets', 'stale')
     # No need to add the transform_result.did column back in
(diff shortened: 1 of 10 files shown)


ponyisi (approved):
Seems OK to me. As discussed on Slack I might have a mild preference for "Trash" as the verb for the action here rather than "Delete" (since nothing is getting deleted from the DB) but not enough to demand a change.

Issue #859, opened 2024-09-26
Archive Transforms


BenGalewsky:
_As a ServiceX user I want to be able to archive old transforms so they no longer clutter my workspace and to free up space_

# Description
This is a wide-ranging change to the way we manage transforms. Introduce the ability to archive an old transform. Doing this will:
1. Remove results from the object store
2. Clear out the detailed database records
3. Mark the transform request as _Archived_ and hide these from the ui

# Approach
The functionality will be implemented in the ServiceX App since this services already has database and object store credentials, configuration, and most of the required python code. There will be two ways of initiating an archive:
1. An internal REST endpoint (only accessible from another pod inside the cluster) to start off a complete data lifecycle workflow.
2. An external, authenticated endpoint which can be used to archive a single transform, which must be owned by the logged in user.

## Archiving a Single Transform
This can be initiated by the user or as part of a bigger data archiving workflow.
1. Set the archived flag on the transform request
2. Delete all of the results from the object store and delete the enclosing bucket
3. Delete the transform_results records associated with the transform

### Deleting Orphaned Datasets
This will be run at the end of the data archiving workflow. It will query datasets for which there are no non-archived transforms. 
1. The dataset_file records will be deleted
2. The dataset record itself will be deleted

### The Data Archive Workflow
This will be activated by a POST on an internal REST endpoint. The payload of the request will include a timestamp. Any transforms earlier that that timestamp will be archived. At the end of that task it will search and destroy orphaned datasets.

The data archive workflow will be triggered by Kubernetes cron job. The worker will simply perform a `curl` operation on the internal workflow REST endpoint.

I considered implementing the cron job inside the app, but since we potentially will have several replicas of the app running, we'd have to figure out how to make this only run once.

PR #915, opened 2024-11-15
Allow users to delete transform requests

BenGalewsky:
First step in #859 - This PR allows users to request deleting a transform from the CLI.

This is accomplished by adding a new boolean to the transform request table, archived. Setting archived to True removes the transform for reports and requests.

When a transform is archived, the transform results table records are deleted and the contents of the bucket that held the transformed files are deleted, along with that bucket.

The action will report an error if you try to delete a transform that is still running


ponyisi (approved):
Looks good

PR #916, opened 2024-11-15
Tell gunicorn to bind to IPv6 addresses

ponyisi:
By default gunicorn was binding only to IPv4 addresses, potentially causing issues with dual-stack k8s clusters. This change will make it bind to both IPv4 and IPv6 local addresses.


BenGalewsky (approved):
Progress!

PR #918, opened 2024-11-17
Protect against null values for transformer.cachePrefix

### LINKED ISSUE #620: Handle null value for transformer.cachePrefix
`transformer.cachePrefix`: null  is causing problems - looks like it populates the flask config with
TRANSFORMER_CACHE_PREFIX =
rather than
TRANSFORMER_CACHE_PREFIX = ""

ponyisi:
Resolves #620


BenGalewsky (approved):
Hooray!

PR #921, opened 2024-11-20
Data Lifecycle Operations

### LINKED ISSUE #859: Archive Transforms

BenGalewsky:
# Problem
Transforms, results, and datasets remain in the database and object store indefinitely.  The MinioCleanup job was an attempt to delete old objects from Minio, but left invalid transform requests in the database. There was no way to delete datasets without cascading foreign key problems.

Resolves #859 

# Approach
All of the management functionality was moved into the App since it already has methods and configuration to manipulate the database as well as the object store.

## Fully Deleting Transform Requests
There was a previous PR to simply allow Transform requests to be archived. We've decided that this is not the right approach for us at the moment since we have no practical use for keeping the archived transforms in the database.

The [first commit](https://github.com/ssl-hep/ServiceX/commit/1dffbd1661f08e3ede313195b1afb67f2c41a9ca) in this PR 
reverts this change and instead makes the delete endpoint actually delete the transform along with the transform results and the objects in the object store.

## The Data Lifecycle Ops Endpoint
The workflow to clean up old data is triggered by a POST on the `/servicex/internal/data-lifecycle` endpoint. This endpoint can only be reached from inside the cluster, so no auth is required to access it.

The POST request has a path argument of `cutoff_timestamp` which is an ISO formatted datetime. Any transforms older than this timestamp will be deleted along with the Transform Results and the transformed files in the object store.

After the old requests are deleted, the workflow garbage collects datasets that are no longer associated with any transform requests. This is now a simple SQL operation since there are no longer any foreign key records to prevent the delete.

## The Kubernetes Job
The final piece of the puzzle is triggering the workflow. This is done with a Kubernetes cron job. When this job runs
it creates a worker pod based on the `python:3.10' docker image. This was chosen as the default since this image underlies 
most of the python pods in the cluster is therefore already pulled on the cluster nodes.

We don't need much from this docker image. All we do is compute a past date that is based on the `retention` value in the helm chart and then pass that date string to the server using a curl POST request.


ponyisi:
Do we have something that deletes rows from the `files` table that are no longer referred to by any transform result?

BenGalewsky:
> Do we have something that deletes rows from the `files` table that are no longer referred to by any transform result?

[This](https://github.com/ssl-hep/ServiceX/blob/44ecd51916b5cd0bfeec536e60c02d1f78823e8c/servicex_app/servicex_app/resources/internal/data_lifecycle_ops.py#L92) is the line

ponyisi:
Ah, great. Do you have a timing for executing that? (I know one of these queries tends to be very slow due to a foreign key consistency check ... but maybe this is crafted in a way that gets around that.)

BenGalewsky:
> Ah, great. Do you have a timing for executing that? (I know one of these queries tends to be very slow due to a foreign key consistency check ... but maybe this is crafted in a way that gets around that.)

I don't have anything useful from my laptop. If you are satisfied with the overall implementation I'd like to merge this, cut an alpha release of the chart and try it out on testing-2. We can make new fixes to the DB if we find performance problems

ponyisi (approved):
Approved, we can resolve timing issues with testing

PR #928, opened 2024-11-25
Fixes to the Delete PR

BenGalewsky:
# Problem
Upon testing the delete transform request change in testing2 environment there are a few issues that have come up.

* There was still a reference to `archived` TransformRequest property in the dashboard query
* The `@ jwt_required` was starting a transaction in the session which messed up explicit transaction management in the delete REST endpoint resource
* I had created a 1.6.0 database migration which added that `archived` flag. We don't want that flag any more and this is not going to be a 1.6.0 release. Deleted that migration file
* I did want to clean up the indexes on `requests` and `files` to make sure deletes could be done efficiently. Added a new 1.5.6 migration to do that


ponyisi (approved):
Urgently needed for testing

PR #930, opened 2024-11-25
Clean up transaction management for file_complete handler

BenGalewsky:
# Problem
The TransformerFileComplete resource handler is the most critical code in the entire stack. It responds to each file in the dataset being transformed, and is responsible for updating the total number of files processed (either successfully, or failure) - these two counters are how we determine if the transform is complete. The endpoint will be hit repeatedly by all of the running transformers. Consequently, database transaction handing is very important to avoid missing files.

The current implementation uses implicit transactions and doesn't manage locks and flushing to the DB. It's possible that this allows for files to be lost during big transform requests.

# Approach
1.  Us the DB session to explicitly manage transactions
2. Update the `record_file_complete` to read the request with `with_for_update` flag set which will lock the record in the db
3. The increments to files are handled in the same transaction
4. To make the file more readable, the `retry` call arguments are captured in a single, new decorator.`file_complete_ops_retry` - 

With this decorator, I ran into a problem with unit tests. Importing the module caused the `current_app.logger` expression to be evaluated. This would throw `RuntimeError: Working outside of application context.` in the unit tests. Worked around this in the decorator to only access that logger if we are inside the flask app

### DIFF
--- a/servicex_app/servicex_app/resources/internal/transformer_file_complete.py
+++ b/servicex_app/servicex_app/resources/internal/transformer_file_complete.py
@@ -26,23 +26,59 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import logging
+import time
 from datetime import datetime, timezone
+from functools import wraps
 from logging import Logger
 
 from flask import request, current_app
+from sqlalchemy.orm import Session
 from tenacity import retry, stop_after_attempt, wait_exponential_jitter, \
     before_sleep_log, after_log
 
 from servicex_app import TransformerManager
 from servicex_app.models import TransformRequest, TransformationResult, db, TransformStatus
 from servicex_app.resources.servicex_resource import ServiceXResource
-import time
+
+
+def file_complete_ops_retry(func):
+    """
+    A decorator that applies a standard retry mechanism for file complete operations.
+
+    This decorator wraps the target function with a retry mechanism using the tenacity library.
+    It attempts to execute the function up to 3 times, with exponential backoff and jitter
+    between retries. The decorator also logs retry attempts and results.
+
+    Note:
+    This decorator is designed to work both within and outside of a Flask application context.
+    It dynamically evaluates the logger at runtime to support usage in various environments,
+    including unit tests.
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+
+        try:
+            from flask import current_app
+            logger = current_app.logger
+        except RuntimeError:
+            logger = logging.getLogger(__name__)
+
+        return retry(
+            stop=stop_after_attempt(3),
+            wait=wait_exponential_jitter(initial=0.1, max=30),
+            before_sleep=before_sleep_log(logger, logging.INFO),
+            after=after_log(logger, logging.INFO)
+        )(func)(*args, **kwargs)
+
+    return wrapper
 
 
 class TransformerFileComplete(ServiceXResource):
     @classmethod
-    def make_api(cls, transformer_manager):
+    def make_api(cls, transformer_manager, logger=None):
         cls.transformer_manager = transformer_manager
+        cls.logger = logger or logging.getLogger(__name__)
         return cls
 
     def put(self, request_id):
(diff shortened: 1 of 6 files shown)


servicex_app/servicex_app/resources/internal/transformer_file_complete.py
```diff
@@ -71,57 +115,55 @@ def put(self, request_id):
...
-        if transform_req is None:
-            msg = f"Request not found with id: '{request_id}'"
-            logger.error(msg, extra={'requestId': request_id})
-            return None
+    @file_complete_ops_retry
+    def record_file_complete(session: Session, logger: Logger, request_id: str,
+                             info: dict[str, str]) -> TransformRequest | None:
 
-        if info['status'] == 'success':
-            TransformRequest.file_transformed_successfully(request_id)
```
ponyisi, line 87: Do these methods (`file_transformed_successfully`, `file_transformed_unsuccessfully`) still exist on `TransformRequest`? If so, should they be removed?

servicex_app/servicex_app/resources/internal/transformer_file_complete.py
```diff
@@ -71,57 +115,55 @@ def put(self, request_id):
...
+                msg = f"Request not found with id: '{request_id}'"
+                logger.error(msg, extra={'requestId': request_id})
+                return None
+
+            if info['status'] == 'success':
+                transform_req.files_completed += 1
+            else:
+                transform_req.files_failed += 1
+
+        session.flush()  # Flush the changes to the database
```
ponyisi: What is the state of the DB if the transaction attempt fails three times? Do we again go out of sync?

What isolation level are we working at? (Are we certainly taking a row lock, or could the transaction fail because it's been updated somewhere else?)

BenGalewsky: There are three operations. One of them is on table of transform results. Each file complete has it's own row, so there is no contention there.

The part that causes us the most concern is the number of files processed column in the TransformTable. That's the one I put the heaviest level of locking on. 

The other interesting one is declaring that the last file has been processed and so we can call the transform complete and shut down the transformers. I would argue that's not catastrophic if it happens twice (It is a big deal when it never happens!)

My biggest concern is deadlock, which is why I make the update of the TransformResult table a different transaction from the update of files remaining. So I think there is no chance for deadlock.

So, but good point about how to handle failures of each transaction/function call in this resource. Problem is, I don't think we have any good rollback options.

1. The update to the files_complete fails: We are in big trouble. There is no way to recover this and the transform will never quit.
2. The update to the transform_result fails... I don't think this will have much impact. As Illija points out I'm not sure who uses this table since we just scan the bucket for files to return in the client.
3. The status update fails. This is also a big deal, but not sure what we could do about it.

I suppose I should put some `try`/`except` blocks in this code to manage this, but as I say I don't know what to do if we catch an exception (and a function fails after three tries)

servicex_app/servicex_app/resources/internal/transformer_file_complete.py
BenGalewsky: Ah, good call.. Removed those methods from TransformRequest

servicex_app/servicex_app/resources/internal/transformer_file_complete.py
ponyisi: OK. I'd say we should try to have an explicit logging of a situation like this. (Some day it might be nice to make an alert on a dashboard or an email?) Is it also possible to mark the transformation failed?

BenGalewsky: This is done [here](https://github.com/ssl-hep/ServiceX/blob/8ca7037838170fe7ead9a2cfdf3e36efa3d0c29f/servicex_app/servicex_app/resources/internal/transformer_file_complete.py#L135) - it's just an Error log with a distinctive message.

transformer_sidecar/src/transformer_sidecar/transformer.py
```diff
@@ -456,7 +463,7 @@ def init(args: Union[Namespace, SimpleNamespace], app: Celery) -> None:
             "--without-mingle",
             "--without-gossip",
             "--without-heartbeat",
-            "--loglevel=warning",
+            "--loglevel=info",
```
ponyisi, line 466: Won't this bring back a lot of uninteresting verbosity?

ponyisi (approved):

transformer_sidecar/src/transformer_sidecar/transformer.py
BenGalewsky: The better to debug with til we are certain this is fixed -- I suppose it could be a helm chart setting, but that's a bit of a chore to work through

PR #931, opened 2024-11-26
Add flower celery web monitor app

BenGalewsky:


servicex_app/migrations/versions/1.5.6_.py
ponyisi, line 1: Is there a potential conflict with this file?

ponyisi (approved):
I feel this PR may be mixing some changes from other PRs as well?

BenGalewsky:
It turns out that flower doesn't work with the mingle protocol turned off as we have done with the transformers. I had hoped it would be useful in debugging our queue issues...

I'm going to withdraw this for now

Issue #922, opened 2024-11-21
Do not report transform complete until upload is complete


ponyisi:
Change the transformer sidecar so that instead of running a background thread that does file uploads (and reporting transform complete when the science payload is done, rather than after the upload of a file is done), that a transformer celery task waits until the file upload is complete before reporting completion.

ponyisi:
@BenGalewsky I think this is resolved now?

PR #932, opened 2024-11-26
Do file upload to object store inside the transformer celery task

ponyisi:
Do not use a separate queue / upload thread; upload results synchronously to a job.


ponyisi (commented):
Resolves #922

BenGalewsky:
At some point we will need to break up transformer.py into smaller files, but not today...

BenGalewsky (approved):
Thanks

PR #933, opened 2024-11-27
Fix logging from sidecar

BenGalewsky:
# Problem
Logging with Celery is complicated. There are several processes in the sidecar app, each with their own logging namespace. Celery wants to manage logging itself, thank you very much.

At some point the sidecar wasn't emitting any logs. Only the object store uploader

# Approach
1. Added a `after_setup_logger.connect` function that is given the logger created by celery so we can configure it correctly
2. Updated `initialize_logging` to accept an existing logger to modify. Make sure it is well and truly INFO level logging
3. Celery packages up print statements and turns them into INFO log lines. The science container communication class was littered with print statements. Turned them into DEBUG log lines


BenGalewsky:
@ponyisi - I confirm that I only use `requestId` tag which is consistent with the important existing queries

ponyisi (approved):

PR #936, opened 2024-12-04
894 objectstore errors

### LINKED ISSUE #894: Failed upload to object store should not mark file as complete
The `upload_file` function in `transformer_sidecar/object_store_manager.py` will silently eat `S3Error` and `MinioException`, but these most likely mean that the result file was *not* in fact uploaded to the object store and the transformation should be marked as a failure (certainly not a success...)

I also don't think there are any retries at all in this particular code path?

BenGalewsky:
# Problem
The object store manager swallows exceptions from uploading the transformed file to the object store. When this happens
the file is not reported at all to the app and the transform will never complete.

Resolves #894 

# Approach
Instead of just logging the exceptions and carrying on, raise a new Exception subclass, ObjectStoreError - catch that in the transformer and post the file-complete message to the server, but with the status set to _failed_

Also took this opportunity to remove the vestigial UploadQueue from the class

### DIFF
--- a/transformer_sidecar/src/transformer_sidecar/object_store_manager.py
+++ b/transformer_sidecar/src/transformer_sidecar/object_store_manager.py
@@ -27,7 +27,13 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import os
 import logging
+import traceback
+
 from minio.error import MinioException, S3Error
+
+
+class ObjectStoreError(Exception):
+    pass
 
 
 class ObjectStoreManager:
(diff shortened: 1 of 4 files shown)


transformer_sidecar/src/transformer_sidecar/object_store_manager.py
```diff
@@ -60,10 +64,12 @@ def upload_file(self, bucket, object_name, path):
                              "requestId": bucket, "object": result.object_name})
         except S3Error:
             self.logger.error("S3Error", exc_info=True)
```
ponyisi, line 68: Does this swallow the message from the underlying `S3Error` or `MinioException`, possibly making it harder to understand what happened?

BenGalewsky: Updated to log the stack trace and then include the original exception in the re-raised exception

ponyisi (approved):

PR #937, opened 2024-12-05
Use login shells in science images

ponyisi:
It can simplify environment variable setup in the science images to invoke bash as a login shell (so `.profile` and friends are read). For example we can automatically set up ATLAS releases in those images.


BenGalewsky (approved):
Yay - I never quite figured out how to do this. Thank you

PR #944, opened 2024-12-18
Pick up the result file size directly from the file before uploading

### LINKED ISSUE #943: File Size Not Always Correctly Reported
# Problem
The `total_bytes` property of the `FileCompleteRecord` comes out of the stats reported by the generated code. This is not always provided, and it's hard to test and fix each of these.

# Approach
The most correct way to report the size of the file is to just ask the operating system. Use the python Path,getsize call to get the correct information to report to the app.

BenGalewsky:
# Problem 
File size reported to file complete endpoint is often zero

Resolves #943 

# Approach
Before uploading the result file from the object store, directly observe the file size using Path.getsize call.


ponyisi (approved):
Looks good

PR #945, opened 2024-12-18
Correctly record the time when a dataset is created and update the last_used time

### LINKED ISSUE #906: Update `last_used` field in Dataset
_As a ServiceX Site administrator I want to know the last time a dataset was used so I can decide when it is safe to delete it_

Right now, we set the `last_used` field to the time when the dataset is created and never change it. Oddly enough, we set the `last_updated` field to 1970-00-00 when the dataset is created. That too is never updated

BenGalewsky:
# Problem
The create time is incorrectly been set to 0 and the last_used time set to the current date/time. The last used time is never updated so it is useless.

Resolves #906 

# Approach
Set the last_updated and last_used fields to the current time when the dataset is created. Update the record to set last_used whenever a cached dataset is used in subsequent transforms

### DIFF
--- a/servicex_app/servicex_app/dataset_manager.py
+++ b/servicex_app/servicex_app/dataset_manager.py
@@ -51,10 +51,11 @@ class DatasetManager:
                  db: SQLAlchemy = None):
         dataset = Dataset.find_by_name(did.full_did)
         if not dataset:
+            dataset_timestamp = datetime.now(tz=timezone.utc)
             dataset = Dataset(
                 name=did.full_did,
-                last_used=datetime.now(tz=timezone.utc),
-                last_updated=datetime.now(tz=timezone.utc),
+                last_used=dataset_timestamp,
+                last_updated=dataset_timestamp,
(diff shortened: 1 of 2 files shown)


servicex_app/servicex_app/dataset_manager.py
```diff
@@ -54,7 +54,7 @@ def from_did(cls, did: DIDParser, logger: Logger, extras: dict[str, str] = None,
             dataset = Dataset(
                 name=did.full_did,
                 last_used=datetime.now(tz=timezone.utc),
-                last_updated=datetime.fromtimestamp(0),
+                last_updated=datetime.now(tz=timezone.utc),
```
ponyisi: Would it be more consistent if `last_updated` were set to `last_used` when the dataset is first created (as opposed to being marginally later, as now)?

BenGalewsky: Nice touch. Fixed that

ponyisi (approved):
Looks good, thanks.

PR #959, opened 2025-02-01
Test harness updates

ponyisi:
Some updates to improve usability. Allow query to be read in from a file, handling for codegen errors, good user for x509 proxy


gordonwatts (approved):
This looks fine to me - but I'm not the original author. It might be @BenGalewsky you need to review this?

BenGalewsky:
This wasn't me either. It was developed by Ketan - it doesn't impact production, so I'm fine with this... Will merge for 1.5.6

PR #970, opened 2025-02-24
Generic OIDC (i.e., Keycloak) implementation for ServiceX authn/authz

ponyisi:
Verified to work with UC AF Keycloak.

Admin must specify the OIDC metadata endpoint URI, as well as a client key and secret. This can be done for Globus as well, so there is no forced change to Keycloak.

The admin can set the `allowAllAfterAuth` config flag to `true`, which will auto-accept any user that is able to pass through the authentication. This will give a "single sign on"-like experience.

When setup, changes lookups of users to be primarily through email, not through the provider `sub` claim. This means that a different underlying provider can be used without messing up the user account.

*The JWT refresh tokens for requests now use email as the identity, not the `sub` claim. Therefore all existing tokens need to be regenerated by the users when this version is installed!*

This does not yet implement any special admin actions, needs further discussion with @ivukotic about how to implement them.


servicex_app/migrations/versions/1.5.7a.py
```diff
@@ -0,0 +1,25 @@
+"""Switch to index on email not sub
+
+Revision ID: v1_5_7a
```
BenGalewsky, line 3: Why isn't this 1.6?

servicex_app/servicex_app/resources/users/token_refresh.py
```diff
@@ -27,19 +27,19 @@
...
 from flask_restful import Resource
 from servicex_app.models import UserModel
 
 
 class TokenRefresh(Resource):
     @jwt_required(refresh=True)
     def post(self):
-        user = UserModel.find_by_sub(get_jwt_identity())
         claims = get_jwt()
+        user = UserModel.find_by_email(claims['sub'])
```
BenGalewsky, line 39: So once the JWT is unpacked, the `sub` is the email address now?

BenGalewsky (changes requested):
One question and one suggestion. This is a wonderful change. I'm so glad it was manageable.

servicex_app/migrations/versions/1.5.7a.py
ponyisi: It could be 1.6. I guess I didn't have a good sense how we want to number these (e.g. if I called it 1.6 then we'd need a 1.6a for anything else we wanted to land for that?) I actually wonder if we should just give these numbers or something instead of trying to suggest they're tied to specific releases.

servicex_app/servicex_app/resources/users/token_refresh.py
ponyisi: Yes. In principle then, if someone changes OIDC providers but the emails don't change, then the tokens remain valid. We could honestly remove the `sub` field from the database but I was reluctant to do that just yet.

BenGalewsky: Just double checking - the 'sub' is the user's email in this jwt?

BenGalewsky (approved):
Looks great... let's ship it!

PR #971, opened 2025-02-26
Eliminate possibility of running arbitrary user-specified science images

ponyisi:
Only take science image specification from codegen. Remove image argument to submission. (Since the frontend doesn't send `null` for optional arguments, this is fully compatible with all requests that didn't try to override the science image, which should be ~ all of them.)

Codegens that want to have multiple science image support should implement an option in their queries and construct the image name themselves.


BenGalewsky (approved):
Nice!

Issue #961, opened 2025-02-03
Finish time on requests for Unknow files


ArturU043:
**Description**
Entering a wrong rucio DID causes the transformation request to enter the complete status with no finish time value.
The web dashboard will pin these requests on top if sorted by Finish(desc) (default)

This could happen if there are typos and it clutters the dashboard when there are multiple

A finish time value and a special status should be attributed to these requests.  

**Example screenshot**

![Image](https://github.com/user-attachments/assets/33bfccfb-6943-49de-b5f7-930d7c286f6e)

PR #972, opened 2025-02-26
Add finish times when transforms run on zero files

ponyisi:
Had previously missed adding finish times for transform status when the jobs are aborted due to having no input files
Should address #961


BenGalewsky (approved):
Sweet

PR #973, opened 2025-02-26
Fix servicex.yaml download

### LINKED ISSUE #926: Fix .service file download
Match 3.0 client 
Set the default backend

ponyisi:
Only give one endpoint for the instance; drop obsolete `codegen` and `return_data` fields. Add `default_endpoint`. Resolves #926


kyungeonchoi (approved):
Finally :)

PR #978, opened 2025-03-03
Show dashboard in user time zone

ponyisi:
Use moment.js to show times in the browser time zone, and indicate to the user which timezone is being used.


BenGalewsky (approved):
This PR demonstrates why none of us want to change the dashboard.  Thank you for your heroism!

PR #979, opened 2025-03-03
Change default dashboard sort order to start (desc)

ponyisi:
The sort order "Finish (desc)" on the dashboard means that requests can get reordered, when reloading the page, in strange and counterintuitive ways. Change the default so that requests are shown by default in the order of submission.


BenGalewsky (approved):
Well, simple enough!

PR #987, opened 2025-03-05
Update to codegen library 1.2

ponyisi:
Will resolve a lot of security warnings, as well as enable codegens to specify science images if they wish


BenGalewsky (approved):
Thanks, dependabot!

PR #992, opened 2025-03-07
endpoint name is the instance name if it's defined

ponyisi:
Make the names in the `servicex.yaml` downloads a little nicer.


BenGalewsky (approved):
Perfect, thank you

Issue #743, opened 2024-04-19
No cancel button for transforms that need a DID finder lookup


gordonwatts:
**Describe the bug**
The first time you run on a data sample that requires a DID finder lookup, you do not get a cancel button - and the transform on the web seems to remain in the `lookup` phase.

**To Reproduce**
1. Submit a transform on a sample that isn't in the cached database in the `servicex_app`
2. Load the webpage dashboard

**Expected behavior**
Expect a cancel button and also number of transformers, and marching ants once the lookup phase is over. In particular, the transform should leave the `lookup` phase.

**Screenshots**
![image](https://github.com/ssl-hep/ServiceX/assets/1778366/671a445b-2777-4b7d-8361-3028f8807e6e)

**Additional context**

kyungeonchoi:
I observed the same behavior and the status left in `Lookup` might be the reason for no cancel button, progress bar, and number of workers.

PR #993, opened 2025-03-08
Allow cancellation of all not-completed requests

ponyisi:
Show the cancel button in the dashboard for all requests that are believed to be in progress (including in the `Lookup`, `Pending Lookup`, and `Submitted` states).


ponyisi:
Fixes #743

BenGalewsky (approved):
Hooray for the cancel button!

PR #995, opened 2025-03-10
Do not login or push to dockerhub if secrets not available in job

ponyisi:
It's important that dependabot be able to build docker images to test that the updated `poetry.lock` files actually work in the container build process. The current setup requires that the Docker Hub secrets be present in the CI job or else things will fail, and dependabot does not have access to these. Use a conditional to avoid pushing the images, and just check we can build them.


kyungeonchoi (approved):
Thanks!

PR #998, opened 2025-03-11
Provide overrides for outdated VOMS servers

ponyisi:
Override the CMS and ATLAS VOMS servers until the upstream RPM packages catch up.


BenGalewsky (approved):
Fine solution to an unfortunate problem...

PR #1014, opened 2025-03-27
Fix regex to capture error message

kyungeonchoi:
Fix regex to capture error message (not the default message)


gordonwatts (approved):

gordonwatts (commented):
I like this! We should be adding more of these, probably!

kyungeonchoi:
Yeah, just a quick fix while I'm preparing tests for TopCP codegen :)

PR #1016, opened 2025-03-28
Update func-adl to fix AST bug

gordonwatts:
Upgrade func-adl to version 3.4.0 and adjust Python version compatibility to 3.9 or higher. Update test dependencies to include 'hatch' and remove the previous version references.


BenGalewsky (approved):
Let's do it

PR #1025, opened 2025-04-04
Enable Pointer Returns in C++

gordonwatts:
This version of `func_adl_xAOD` enables C++ pointers to be returned from injected C++ code.

This is especially helpful when dealing with ElementLink's that have to by dynamically cast to other object types (knarly, dude).

This is suitable for inclusing in a release as soon as it has passed review and tests.

### DIFF
--- a/code_generator_funcadl_xAOD/pyproject.toml
+++ b/code_generator_funcadl_xAOD/pyproject.toml
@@ -8,7 +8,7 @@ packages = [{ include = "code_generator_funcadl_xaod" }]
 
 [tool.poetry.dependencies]
 python = "~3.10"
-func-adl-xAOD = "2.2.0"
+func-adl-xAOD = "^2.2.0"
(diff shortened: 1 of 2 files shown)


code_generator_funcadl_xAOD/pyproject.toml
```diff
@@ -8,7 +8,7 @@ packages = [{ include = "code_generator_funcadl_xaod" }]
 
 [tool.poetry.dependencies]
 python = "~3.10"
-func-adl-xAOD = "2.2.0"
+func-adl-xAOD = "^2.2.0"
```
ponyisi, line 11: I'm a bit confused here - doesn't this syntax allow 2.2.0 to be used? Don't you actually want ^2.2.1 ? (or >=2.2.1)

BenGalewsky (approved):
Let's release and verify

Issue #999, opened 2025-03-16
RNtuple output support


ponyisi:
Support RNtuple as an output format when uproot has full writing capability.

* uproot-raw, func_adl uproot transformers can support RNtuple via uproot write API,
* sidecar data handler can implement awkward -> RNtuple,
* specialty transformers (e.g. TopCP) need this supported in their backends and are out of scope.

ponyisi:
RNtuple is supported as an `uproot-raw` output since 1.6.1 + uproot science container 5.6.1

ponyisi:
@gordonwatts is this something you can define how to implement for func_adl uproot?

gordonwatts:
Could we just use the converter? AnalysisBase and friends don't support it yet, and even if they did, it would only be the case for things going forward.

ponyisi:
All known edge cases now handled by sidecar

PR #1026, opened 2025-04-05
Support RNTuple as possible output format

ponyisi:
* Support RNTuple as output format for the `uproot-raw` and `func_adl_uproot` transformers
  * Add necessary enums; update uproot-raw transformer
* `uproot` 5.6 doesn't yet implement compression when writing RNTuple, so this option is not yet recommended for production.

### DIFF
--- a/code_generator_raw_uproot/servicex/templates/transform_single_file.py
+++ b/code_generator_raw_uproot/servicex/templates/transform_single_file.py
@@ -26,7 +26,7 @@ def transform_single_file(file_path: str, output_path: Path, output_format: str)
 
         ttime = time.time()
 
-        if output_format == 'root-file':
+        if output_format in ('root-file', 'root-rntuple'):
             import uproot
             etime = time.time()
             # opening the file with open() is a workaround for a bug handling multiple colons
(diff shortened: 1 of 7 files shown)


ponyisi:
Good to go here I think for RNTuple output (see #999). I do not recommend rntuple yet (uproot does not yet have compression enabled for the output in 5.6.0) but the code paths technically work and produce correct output.

code_generator_raw_uproot/servicex/templates/transform_single_file.py
```diff
@@ -26,20 +26,34 @@ def transform_single_file(file_path: str, output_path: Path, output_format: str)
 
         ttime = time.time()
 
-        if output_format == 'root-file':
+        if output_format in ('root-file', 'root-rntuple'):
```
gordonwatts, line 29: Why not put this as another top level output option? I get they are both `uproot`, but the actual code is quite different, and now this logic is buried.

code_generator_raw_uproot/servicex/templates/transform_single_file.py
```diff
@@ -26,20 +26,34 @@ def transform_single_file(file_path: str, output_path: Path, output_format: str)
...
+                            else:
+                                writer.mktree(k, dict(zip(v[1].form.columns(),
+                                                          v[1].form.column_types())))
+                        else:  # RNTuple
+                            if v[0] is not None:
+                                # Work around a limitation in uproot 5.6.0
+                                # If a cut is specified, we'll get ListArrays which can't be
+                                # written via uproot. Convert them to ListOffsetArrays
+                                # Assume the ListArrays are only at top level
+                                warr = ak.zip({_: v[0][_].layout.to_ListOffsetArray64()
```
gordonwatts, line 50: Does this imply a limitation on what we can do? If so, add to the PR comments in the title so that it will automatically show up in the change log?

gordonwatts (approved):
Huh - copilot is now reviewing our PRs!

Ok - this looks fine - My  only comments are style comments (nested if statements, etc.). I also cleaned up the PR text because it is often used to generate change logs. I'll mark approve, but you might made modifications regardless.

code_generator_raw_uproot/servicex/templates/transform_single_file.py
ponyisi: It assumes `uproot.iterate()` outputs in a specific format (which is determined by `uproot` internals, which depend in some complex way on the user query). Whether this constrains possible queries is not yet understood, so we're not in a position to give guidance here yet.

code_generator_raw_uproot/servicex/templates/transform_single_file.py
gordonwatts: Discussed this privately.

ponyisi:
I should note that the next `uproot` release will remove the need for funny `ListArray` manipulation and will compress the `RNTuple`s, so it makes sense to hold off on this PR until we get the new version. (The code here should still work with the new release but will have code branches that are no longer relevant.)

BenGalewsky (approved):
Excellent

PR #1027, opened 2025-04-06
Ignore missing trees in uproot-raw transformer by default

ponyisi:
By default, now, if a requested source tree for a query is missing in an input file, the uproot-raw transformer will ignore it instead of raising an exception. This improves compatibility with ATLAS TopCPToolkit output. The old behavior can be restored by adding `fail_on_missing_trees: true` to the query JSON.


BenGalewsky (approved):
Thank you

PR #1029, opened 2025-04-08
Set uproot science image to 5.5.1

ponyisi:
Give a specific version for the uproot science image so deployments don't pick up every change to the uproot5 branch in an uncontrolled way


BenGalewsky (approved):
Sure

PR #1030, opened 2025-04-08
Store transform events and size in DB

ponyisi:
Store the `total_events` and `total_bytes` fields in the database.

Right now the information will be updated on the completion of each file. I don't know if this is necessarily useful or if we should wait until transform completion, although updating as we go makes more sense from a user display point of view (if we update only at the end of the job we should probably hide the relevant fields from the user-facing page until the transform is done).


gordonwatts:
I love this - I would approve it if I was reviewing it, but feels like something Ben should look at since the interactions might be subtle.

BenGalewsky (approved):
Will be curious to see which transformers support this, but this is good

PR #1032, opened 2025-04-18
New version of func-adl-xaod C++ Codegen

gordonwatts:
* Nested arrays sometimes had `push_back` placed at the incorrect scope
* Error messages during codegen improved


BenGalewsky (approved):
Opaque and wonderful

PR #1036, opened 2025-04-23
Fix silly bug in min workers in hpa request

ponyisi:
Don't allow min workers to exceed max workers requested for very small datasets


BenGalewsky (approved):
Hee hee

Issue #861, opened 2024-09-26
Specify resources for all objects in helm chart


BenGalewsky:
_As an Analysis Facility admin I need all pods in ServiceX to specify their resource requirements so I can schedule them on contested nodes_

# Description
Add CPU and Memory requests to all services in the helm chart. Insure that the TransformerManager sets them correctly for the transformer pods too

PR #1037, opened 2025-04-23
Set transformer memory resource limit and make configurable

ponyisi:
Some k8s installations may have a very low default limit for pod memory, which leads to transformers dying with no warning due to memory exhaustion. Ask for 2GB by default and allow sites to configure this. (It's the same for all transformers; I suppose in the future we might revisit that.)


ponyisi:
Addresses one component of #861

BenGalewsky (approved):
Ah, thank you

PR #1041, opened 2025-04-23
Eliminate three unnecessary log lines per worker

ponyisi:
Since we init an empty `TransformerStats` in each sidecar, we were getting three log entries from the print statement which were also quite confusing (i.e. "File does not exist:" and "None", with an empty string also). Clean this all up, backstop with a dummy message if there _is_ an issue.


BenGalewsky (approved):
Listen to the silence....

PR #1042, opened 2025-04-23
Upgrade TopCP codegen library dependencies

ponyisi:
Bump codegen library version to 1.2, enables science image tag choice and resolves security alerts


kyungeonchoi (approved):
Thanks! Next TopCP codegen PR will benefit from this!

PR #1044, opened 2025-04-24
Bump uproot science image to 5.6.0

ponyisi:
Better version to invoke (in particular we can get the ARM build)


BenGalewsky (approved):
Forward!

PR #1046, opened 2025-04-25
Add missing quotes that broke 1.6.1-RC2

ponyisi:
Needed because the memory limit is a string


BenGalewsky (approved):
What a difference a pair of quotes make...

PR #1049, opened 2025-04-29
remove s3 bucket polling

### LINKED ISSUE #1006: Don't Poll S3 for Bucket Contents
Instead of polling S3 bucket for contents create a new App endpoint that returns all files since a timestamp:

1. Add DB generated timestamp for each result file
2. Include timestamp in the response
3. Create endpoint that returns files since timestamp

MattShirley:
Added in preparation for front end changes per https://github.com/ssl-hep/ServiceX/issues/1006


servicex_app/servicex_app/resources/internal/get_dataset_files_since.py
```diff
@@ -0,0 +1,42 @@
...
+        if args.get('cutoff'):
+            cutoff_str = args['cutoff']
+
+            try:
+                if cutoff_str.endswith('Z'):
+                    cutoff_str = cutoff_str[:-1] + '+00:00'
+
+                cutoff_datetime = datetime.fromisoformat(cutoff_str)
+
+                files_query = files_query.filter(DatasetFile.created_at >= cutoff_datetime)
```
BenGalewsky: I think this needs to be `>` since the time you would pass into this call is the timestamp for the last file the client processed

servicex_app/servicex_app/models.py
```diff
@@ -463,14 +463,18 @@ class DatasetFile(db.Model):
     file_events = db.Column(db.BigInteger, nullable=True)
     paths = db.Column(db.Text(), unique=False, nullable=False)
     dataset = relationship("Dataset", back_populates="files")
+    created_at = db.Column(DateTime, default=func.now())
+    updated_at = db.Column(DateTime, default=func.now(), onupdate=func.now())
```
BenGalewsky: Not sure when we'd ever use the `updated_at` column. I think files are written once and that's the end of it

ponyisi:
Looks good to me. I guess by default the index will be a B-tree and will be useful for the > operator?

ponyisi:
Another question, for broader discussion - is it finally time for us to introduce a "capabilities" endpoint? In this case we could use it so the client could verify that this new endpoint is available and switch behavior appropriately. (Mostly I am starting to worry about different sites running different versions of the server software but all users picking up the latest versions of the clients.)

MattShirley:
> Looks good to me. I guess by default the index will be a B-tree and will be useful for the > operator?

Yeah, any index will be B-tree by default and I think it's good practice to over-index anything that might be used to filter values.

MattShirley:
> Another question, for broader discussion - is it finally time for us to introduce a "capabilities" endpoint? In this case we could use it so the client could verify that this new endpoint is available and switch behavior appropriately. (Mostly I am starting to worry about different sites running different versions of the server software but all users picking up the latest versions of the clients.)

Can these capabilities be reduced down to version numbers? If so, you could set the client version as a HTTP header, version the endpoints, and create a middleware to fail any request without a sufficient client version for the endpoint.


I spent some time hacking with Claude and think an interface like this should be possible:

```
class Endpoint(VersionedResource):
    pass

@Endpoint.require_client(3, 1, 1)
class EndpointV1_0_0(Endpoint):
    def get(self):
        return {"server_version": "v1.0.0"}

@Endpoint.require_client(3, 2, 0)
class EndpointV1_1_0(Endpoint):
    def get(self):
        return {"server_version": "v1.1.0"}
        
api.add_resource(Endpoint, '/my_versioned_endpoint')
```

MattShirley:
@ponyisi @BenGalewsky I made major changes to this PR, please take a look. I noticed that files were available in the `files` table as soon as the request was made (and before results were available) while `transformation_results` streamed into our database as they became available. So now the endpoint checks transformation_results, which will be polled by the client. Please let me know if this change makes sense and whether we already have an endpoint that offers similar functionality.

MattShirley:
@ponyisi @BenGalewsky added the feedback from Friday, including the cutoff/begin_at value to the endpoint.

servicex_app/servicex_app/resources/transformation/results.py
```diff
@@ -0,0 +1,43 @@
...
+        parser.add_argument(
+            'begin_at',
+            type=str,
+            required=False,
+            location='args'
+        )
+
+        args = parser.parse_args()
+
+        transform_result_query = TransformationResult.query.filter_by(request_id=request_id)
```
ponyisi, line 29: do we have an index on `request_id` ?

BenGalewsky:
Please contribute to the quality of our codebase with unit tests for this new endpoint

servicex_app/servicex_app/resources/transformation/results.py
MattShirley: We do not. I will add one.

MattShirley:
The feature flag has been added and tests updated/added. This is ready for release from my side.

BenGalewsky (approved):
Thank you! This is such a great contribution

PR #1051, opened 2025-05-02
improve local dev for server side development

MattShirley:
How to use:

1. Replace your local helm chart with the one available in this PR for: `helm/servicex/templates/app/deployment.yaml`
2. Update your local helm values file to include:
```app:
  environment: dev
  mountLocal: true
  reload: true
```
3. Install Overmind. You can view installation documentation here: https://github.com/DarthSim/overmind. While the instructions include installing tmux, you will also need to install ruby if not installed: `brew install ruby` or `sudo apt install ruby` or your OS's equivalent. Note that when installing Overmind via gem, you may need to install with the `--local-install` argument. i.e. `gem install overmind --user-install`. If you enable this option, Overmind will be installed in your home directory. You will need to locate it and make sure it is available in your path.
4. Navigate to your ServiceX checkout. You will need to set your ServiceX directory, your helm chart directory, and your helm values file as environment variables (the values file can be either absolute or relative to $CHART_DIR). My command is:
```LOCAL_DIR=/home/matt/Projects/ServiceX CHART_DIR=/home/matt/servicex VALUES_FILE=local-values.yaml overmind start```
5. Since that command is a bit unwieldy, I suggest adding aliases inside your `~/.bashrc` file.
```
alias servicex-up='cd ~/Projects/ServiceX && LOCAL_DIR=/home/matt/Projects/ServiceX VALUES_FILE=local-values.yaml CHART_DIR=/home/matt/servicex overmind start'
alias servicex-down='helm delete servicex'
alias servicex-start='servicex-up; servicex-down;'
```

Open a new terminal or execute `source ~/.bashrc` to reload your environment. Now you can start servicex with `servicex-start`. Killing the process will also uninstall the helm deployment (but minikube will still be running).

Once complete, this PR introduces two changes:

1. Changes to your local ServiceX checkout will be immediately reflected in your running app pods and gunicorn workers reloaded.
2. The minikube mount process and separate port forward processes are collected into a single Procfile and executed via Overmind.

overmind

### CODE REVIEW

File: servicex_app/servicex_app/transformer_manager.py [hunk 1]
```diff
@@ -41,288 +41,379 @@ class TransformerManager:
...
         volumes.append(
             client.V1Volume(
-                name='sidecar-volume',
-                empty_dir=client.V1EmptyDirVolumeSource())
+                name="sidecar-volume", empty_dir=client.V1EmptyDirVolumeSource()
+            )
         )
 
+        # Only mount local volumes in development environment with mountLocal enabled
+        if current_app.config.get(
```
MattShirley, line 124: This conditional block contains the new additions of this PR (the rest are because I ran `black` against the file).

ponyisi (approved):

ponyisi:
@MattShirley at some point let's add the documentation in the PR description to the docs directory

PR #1053, opened 2025-05-08
More general python codegen

ponyisi:
Allow the function signature
```
run_query(file_path, output_path: Path)
```
in addition to
```
run_query(file_path)
```
so Python transformers can create arbitrary ROOT file output.

### DIFF
--- a/code_generator_python/python_code_generator/templates/transform_single_file.py
+++ b/code_generator_python/python_code_generator/templates/transform_single_file.py
@@ -2,8 +2,10 @@ import os
 import sys
 import time
 from pathlib import Path
+import inspect
 import generated_transformer
-instance = os.environ.get('INSTANCE_NAME', 'Unknown')
+
+instance = os.environ.get("INSTANCE_NAME", "Unknown")
 default_tree_name = "servicex"
 default_branch_name = "branch"
 
@@ -18,18 +20,23 @@ def transform_single_file(file_path: str, output_path: Path, output_format: str)
     try:
         stime = time.time()
 
-        # We first see if the function has the signature to directly write output
+        # We first see if the function takes two parameters; if so we assume the second
+        # will be interpreted as the file name for the output.
         # If it doesn't, then we assume it's giving us back awkward array results
-        try:
+
+        provided_signature = inspect.signature(generated_transformer.run_query)
+        if len(provided_signature.parameters) == 2:
             generated_transformer.run_query(file_path, str(output_path))
             if not output_path.exists():
-                raise RuntimeError("Transformation did not produce expected output file "
-                                   f"{output_path}")
+                raise RuntimeError(
+                    "Transformation did not produce expected output file "
+                    f"{output_path}"
+                )
             ttime = time.time()
             etime = time.time()
             wtime = time.time()
             total_events = 0
-        except AttributeError:
+        else:
             import awkward as ak
             import uproot
             import pyarrow.parquet as pq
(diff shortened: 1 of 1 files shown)


BenGalewsky:
I don't understand how this works? How does the sidecar know to pick up the user specified file?

ponyisi:
@BenGalewsky The sidecar tells the codegen shim what file name to produce. This PR introduces a route where user code can just directly produce this file, instead of returning an awkward array which the codegen shim then writes to that file.

code_generator_python/python_code_generator/templates/transform_single_file.py
```diff
@@ -22,58 +18,74 @@ def transform_single_file(file_path: str, output_path: Path, output_format: str)
...
+        except AttributeError:
+            import awkward as ak
+            import uproot
+            import pyarrow.parquet as pq
+            import numpy as np
+
+            output = generated_transformer.run_query(file_path)
+
+            ttime = time.time()
+            if output_format == 'root-file':
```
gordonwatts: What about parquet, or all the other output formats that are allowed?

gordonwatts (commented):
The if statements here are getting quite complex. I don't think we have any tests for this. Is this getting to the point we need those?

code_generator_python/python_code_generator/templates/transform_single_file.py
ponyisi: Well, this is why we should really have a "raw" option for the output format. The most obvious need for this code path is running `RDataFrame` transformations, where the output is in fact going to be a ROOT file. 

Conceptually I guess the big question is whether we want to try to share the Python transformer code between the uproot science image (where a "translate awkward output to root/parquet" step is natural) and the C++ ROOT image (where in the end the user writing the Python code is responsible for the output ... unless we ask them to return an RDF object for snapshotting, or something - note this isn't an entirely implausible route, it's possible to do RDF -> Awkward -> Parquet).

gordonwatts: Ah, that is a good point. In my mind I was thinking this was common code.

We have a capabilities matrix, right - so the user can't request an unsupported format? Or if you request parquet from somethign that doesn't support it you get back root? I can't remember how we handle that now (in the old days, we'd just make best effort to pay attention to that request).

ponyisi: @gordonwatts It's actually a lot sillier. Basically, the format request is a hint as far as ServiceX is concerned; *if* the transformer says it can only make root, and the user requests parquet, the sidecar will do a translation, but otherwise it is ignored by the common ServiceX infrastructure and is only interpreted by the individual transformers. (So in particular, if we had something that could only output parquet but we requested root, we would just get the parquet output if the transformer didn't put in its own protections.)

In the case of the old python transformers, we could still play this game somewhat sensibly since the transformation returned an awkward array and we could then write it as root or parquet. But if we allow the function to directly write its output (and that's really the only sensible option in the RDF world, unless we demand it return a `RDataFrame` ?) then we no longer have control over what, in fact, gets made, and we should be very careful about implying that the data type flag actually means anything.

Strictly speaking we could restrict the capabilities of the user code here quite a bit and *force* people to use `RDataFrame` in their Python code (as opposed to the current setup, which allows any valid PyROOT code) - thing is, at the moment the code has all capabilities of the uproot-raw transformer other than writing parquet, while we'd have to do some additional interface design if we wanted to keep all capabilities but restrict the users to RDF). Maybe best to discuss on Thursday.

MattShirley (approved):

code_generator_python/python_code_generator/templates/transform_single_file.py
```diff
@@ -22,58 +18,74 @@ def transform_single_file(file_path: str, output_path: Path, output_format: str)
...
         stime = time.time()
 
-        output = generated_transformer.run_query(file_path)
-
-        ttime = time.time()
-
-        if output_format == 'root-file':
+        # We first see if the function has the signature to directly write output
+        # If it doesn't, then we assume it's giving us back awkward array results
+        try:
```
BenGalewsky: This is getting so ugly... could you please use the [`inspect.signature`](https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object) instead of this exception to make it a little more obvious what the intention is

ponyisi: ok

code_generator_python/python_code_generator/templates/transform_single_file.py
ponyisi: Anyway to be clear the new code path just ignores the `output_format` attribute completely - the particular code you highlighted comes into play only if the user-provided code returns an awkward array, in which case we need to decide what to do with it.

BenGalewsky (approved):
I'm happy with this now....

PR #1054, opened 2025-05-10
Change axis default on a number of functions in uproot-raw codegen

ponyisi:
We make many `awkward` functions available when writing cuts and expressions for uproot. A lot of these default to operating on the entire array at once (`axis=0`) in ways that make no sense for an event-by-event cut mask. In general this default causes a lot of errors. We will override those defaults here (we believe the defaults are so incorrect that nobody with correct code will be affected by this). The axis can still be overridden if people want. The list of functions with changed defaults will be given in the frontend documentation.

### DIFF
--- a/code_generator_raw_uproot/servicex/raw_uproot_code_generator/request_translator.py
+++ b/code_generator_raw_uproot/servicex/raw_uproot_code_generator/request_translator.py
@@ -28,8 +28,11 @@
 import os
 import shutil
 
-from servicex_codegen.code_generator import CodeGenerator, GeneratedFileResult, \
-    GenerateCodeException
+from servicex_codegen.code_generator import (
+    CodeGenerator,
+    GeneratedFileResult,
+    GenerateCodeException,
+)
 
 
 class RawUprootTranslator(CodeGenerator):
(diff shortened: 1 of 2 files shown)


gordonwatts (commented):
I really don't like this. I get why you are doing this. But this is magic and now behaves differently than regular code. So what users learn will break here.

btw - if axis argument is specified twice will this work?

ponyisi:
It will work if the user specifies `axis` again. (It doesn't work if a user specifies it as a positional argument instead of a keyword one but this is a very rare pattern.)

I understand why this would make one uncomfortable but the fact is that essentially every new user I've seen work with the `uproot-raw` transformer makes the error of not specifying `axis=1` in a relevant place and unless they're quite competent with `awkward` they don't figure it out on their own. When awkward specifies `axis=0` as a default this is _never_ the behavior you want here (unless you can come up with a circumstance where we want queries to correlate across events ...). The biggest likelihood of causing problems, I would guess, is actually that some inexperienced users come up with some cut expression that works in `uproot-raw` and not when they run awkward locally.

The equivalent `RDataFrame` expressions don't need this kind of silliness to work correctly...

ponyisi:
I suppose we could make it an option in the `uproot-raw` request (though I would still want the default to be the override).

gordonwatts:
> I suppose we could make it an option in the `uproot-raw` request (though I would still want the default to be the override).

Ok - that makes sense to me. In the documentation could we also make it very clear?

I guess the fundamental problem is that `awkard` was designed for multi-dimensional arrays. `RDF` was defined for HEP events of multi-dimensional arrays. `func-adl` does this too, implicitly.

BenGalewsky:
@ponyisi please update the docs as part of this PR

ponyisi:
@BenGalewsky the docs will go into the servicex_frontend anyway, no?

BenGalewsky:
> @BenGalewsky the docs will go into the servicex_frontend anyway, no?

Ah, right

ponyisi:
@BenGalewsky Documentation in https://github.com/ssl-hep/ServiceX_frontend/pull/606 . I also added a query option to allow the user to revert this change, as discussed in the Thursday meeting. Both modes have been tested on testing-3.

kyungeonchoi (approved):
Looks good to me. Thanks!

kyungeonchoi:
> @BenGalewsky Documentation in [ssl-hep/ServiceX_frontend#606](https://github.com/ssl-hep/ServiceX_frontend/pull/606) . I also added a query option to allow the user to revert this change, as discussed in the Thursday meeting. Both modes have been tested on testing-3.

It would be also nice to specify the ServiceX backend version with this change in the client documentation [ssl-hep/ServiceX_frontend#606](https://github.com/ssl-hep/ServiceX_frontend/pull/606)

PR #1056, opened 2025-05-15
add meta resource that lists all the other available resources available from the server

MattShirley:


ponyisi:
I guess the idea here is then that we don't enumerate "capabilities" but rather we assume a capability is related to the existence of a specific endpoint? (Mostly, my concern is that we might e.g. use an existing endpoint but add a new feature to it, so the mere existence of the endpoint doesn't tell you what you really want to know.)

ponyisi:
(Also, could you add more info to the description so we can track the motivation for the PR a bit better?)

MattShirley:
> I guess the idea here is then that we don't enumerate "capabilities" but rather we assume a capability is related to the existence of a specific endpoint? (Mostly, my concern is that we might e.g. use an existing endpoint but add a new feature to it, so the mere existence of the endpoint doesn't tell you what you really want to know.)

I think we still need to define what a capability is. If we just want to use these as feature flags, I think the better way to go is to just publish a server version number and allow the client to define feature flags as needed. For example:

```
server_version = semantic_version.Version(self.servicex.get_server_version())
transformation_results_enabled = server_version > semantic_version.Version('1.6.1')
```

Having an end point available to return dynamic info only makes sense to me if a ServiceX instance could have a dynamic configuration (i.e. something like plugins, multiple cloud vendor support for example). Do we have any plans to introduce such configuration? Otherwise I'm not seeing a use case for adding more complexity.

ponyisi:
I wasn't thinking of an endpoint with dynamic info (although that is possibly an interesting use case); at the moment I really was just envisioning an endpoint that returned a **hard-coded** JSON string, which would be extended as part of the PR that introduced a new feature we wanted to declare. There need be nothing dynamic at run time at all.

I still think this is better than "send a version number and translate it to features on the client side" because the latter still requires a new version endpoint (which may be a little tricky to keep in sync properly, especially if we take the numbers from the chart), and now adds the complication of maintaining a parallel table in the client, doing to logic to figure out the ordering of version numbers, etc. Why not just return the result of that lookup directly from the server, where it's easy to be authoritative?

PR #1064, opened 2025-05-29
Add capabilities reporting feature

ponyisi:
Extend the "Info" (/servicex) endpoint return value to include a static list of capabilities that the client can test against. For now keep it blank, but this can be updated as features are added.


MattShirley (approved):

MattShirley:
Looks good! Thanks for catching this

PR #1067, opened 2025-06-03
Configurable compression algorithm and level for Uproot transformers

kondratyevd:
This PR adds parameters to control compression algorithm and level, for now only in uproot-raw transformer. 
The new parameters `result-compression-algorithm` and `result-compression-level` are propagated similarly to `result-format`.


BenGalewsky:
Compression differs for different file formats. 

Three alternatives:
1. Make it a code gen default via `values.yaml`
2. Add it to each query syntax to allow this to be passed in
3. Keep it in the request level

BenGalewsky:
Temporarily avoid this issue via #1081

BenGalewsky:
Superseded by #1091

PR #1088, opened 2025-06-20
Enable rdataframe codegen

ponyisi:
Switches on a "rdataframe" codegen, which uses the Python codegen but runs in a C++ ROOT-based science image


BenGalewsky (approved):
Sure, make it so

PR #1092, opened 2025-06-24
add proper s3 object sourcing

MattShirley:
Corrects S3 object name loading from https://github.com/ssl-hep/ServiceX/pull/1049 (which incorrectly sourced it from file paths)


BenGalewsky (approved):
Ok!

PR #1093, opened 2025-06-25
title name of request bump to 512 chars

ArturU043:
In relation with [PR](https://github.com/ssl-hep/ServiceX_frontend/pull/599) from ServiceX_analysis_utils


BenGalewsky (approved):
That does it!

ponyisi:
If you don't mind, I'd like to add a capability report here, so the client can check if this extension is implemented.

MattShirley:
@ponyisi does it need to be handled by the front end or do you just want a key added to the list? Do you have a key name in mind?

PR #1094, opened 2025-06-25
Add R25 science image and codegen

gordonwatts:
Introduce support for the R25 science image and update code generation configurations accordingly by updating `values.yaml`


ponyisi (approved):
One minor change, approved

PR #1095, opened 2025-06-25
uproot 5.6.2 and ROOT 6.36.00 images

ponyisi:
Update science images for uproot and C++ ROOT


gordonwatts (approved):
Changes look great! pull policy - I always forget that.

PR #1101, opened 2025-07-02
add archived bitnami shell image to default values.yaml

MattShirley:
We are currently using older versions of several bitnami images (rabbitmq, postgresql, minio) that use a deprecated image (`bitnami/bitnami-shell`) to manage volume permissions. The deprecated image can be accessed using `bitnami/bitnami-shell-archived` and a replacement image is provided by `bitnami/os-shell` (available in newer versions of the chart repositories).

This PR sets the relevant pods to use the archived version of the deprecated bitnami image. A separate issue will be made to create a repository upgrade plan which will remove the need to specify this image explicitly in our app's values.yaml.


BenGalewsky (approved):
Thanks. this should work

PR #1103, opened 2025-07-03
remove minio.apiIngress.enabled from values.yaml

MattShirley:
Removes the `minio.apiIngress.enabled` value from the default `valules.yaml`. (It is being set to false, which is the default value in the minio chart.) This change was erroneously made to resolve a helm linting error as part of https://github.com/ssl-hep/ServiceX/pull/1101.

### DIFF
--- a/helm/servicex/values.yaml
+++ b/helm/servicex/values.yaml
@@ -143,8 +143,6 @@ minio:
   volumePermissions:
     image:
       repository: bitnami/bitnami-shell-archived
-  apiIngress:
-    enabled: false
   auth:
     rootPassword: leftfoot1
     rootUser: miniouser


ponyisi:
Hi @MattShirley , so we should give this another go soon. Is the current situation harmless? I note that this parameter gets renamed in the newer versions of the minio chart (https://artifacthub.io/packages/helm/bitnami/minio 's upgrade section)

MattShirley:
@ponyisi yes, I think we should add this PR and remove apiIngress from values.yaml. It's passing the linter, which was the only reason apiIngress was added in the first place.

ponyisi (approved):

PR #1104, opened 2025-07-04
Extend sample title field and add corresponding capability

ponyisi:
In discussion with Gordon, it seems that we want "Rucio dataset name + metadata" in the sample title, so 512 characters is in principle not enough. In practice in my opinion the only reason to put a length in this field is to prevent malicious injection of bad data of indefinite length. I propose to raise the limit to 10240 (far more than anything reasonable but not so long as to allow compromising junk). In addition I have added a capability report so the client knows how long of a limit is supported.

### DIFF
--- a/servicex_app/servicex_app_test/resources/test_servicex_info.py
+++ b/servicex_app/servicex_app_test/resources/test_servicex_info.py
@@ -30,18 +30,19 @@ from servicex_app_test.resource_test_base import ResourceTestBase
 
 class TestServicexInfo(ResourceTestBase):
     def test_get_info(self, client, mock_app_version):
-        response = client.get('/servicex')
+        response = client.get("/servicex")
         assert response.status_code == 200
         print(response.json)
         assert response.json == {
-                                 'app-version': '3.14.15',
-                                 'code-gen-image': {
-                                    'atlasxaod': 'sslhep/servicex_code_gen_func_adl_xaod:develop',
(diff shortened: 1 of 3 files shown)


servicex_app/servicex_app_test/resources/test_servicex_info.py
```diff
@@ -30,18 +30,19 @@
...
+            "app-version": "3.14.15",
+            "code-gen-image": {
+                "atlasxaod": "sslhep/servicex_code_gen_func_adl_xaod:develop",
+                "cms": "sslhep/servicex_code_gen_cms_aod:develop",
+                "python": "sslhep/servicex_code_gen_python:develop",
+                "uproot": "sslhep/servicex_code_gen_func_adl_uproot:develop",
+            },
+            "capabilities": [
+                "poll_local_transformation_results",
+                "long_sample_titles_10240",
```
gordonwatts, line 46: Just to make sure - that `0` at the end is intentional? Seems... big...

gordonwatts (commented):
Looks great - just the zero... Making should be there.

servicex_app/servicex_app_test/resources/test_servicex_info.py
ponyisi: Intentional. Bytes are cheap, this should never trigger unless there's something really bad with the submission.

BenGalewsky (approved):
Great!

PR #1116, opened 2025-07-15
Add pre-commit configuration

ponyisi:
Trigger the same formatting requirements as for the frontend


gordonwatts (approved):
Oh yes - better get this in quickly before it causes unending problems with other PRs!

PR #1120, opened 2025-07-19
Add `assert_` to called_once and other checks

gordonwatts:
* convert `called_once` to `assert_called_once`. The former was not actually checking anything.
* Simple fixes to repair tests that were then failing.


gordonwatts (commented):
Many of the test failures look like because the `db` mock isn't inserted properly. As a result, the real `db` is called, rather than the mock, so when you test the mock, it looks like it was never called (thought it was - you can set a break point).

gordonwatts (commented):
There are no reviews, but I think all of us have checked this. Can we merge this yet?

ponyisi (approved):

PR #1133, opened 2025-08-02
Allow list of available xcache servers to be downloaded

ponyisi:
Add `values.yaml` options to allow the specification of a "Virtual Placement Service" site which gives a list of xcache servers to use. In principle this includes also a liveness test, so servers that go down should be removed dynamically from the list we use for transforms. This will be checked if the list hasn't been updated within a certain interval, by default 30 minutes. This new option can fully replace the server list given by `cachePrefix`.

This is not my favorite way to implement this, but since gunicorn spawns multiple workers which have distinct configuration data, this is the simplest way of ensuring that all app instances will eventually update. It would conceptually be cleaner to store global configuration data that might be updated in a single location (Redis?) but that's a much bigger refactoring.


servicex_app/servicex_app/transformer_manager.py
```diff
@@ -410,6 +414,31 @@ def persistent_volume_claim_exists(self, claim_name, namespace):
...
+        if not (thissite := current_app.config.get("TRANSFORMER_CACHE_VPS_SITE", None)):
+            return
+
+        lastchecktime = current_app.config.get("TRANSFORMER_CACHE_VPS_LASTCHECK", 0)
+        ttl = current_app.config.get("TRANSFORMER_CACHE_VPS_INTERVAL", 30 * 60)
+        if (now := time.time()) - lastchecktime > ttl:
+            current_app.config["TRANSFORMER_CACHE_VPS_LASTCHECK"] = now
+            vps_server = current_app.config["TRANSFORMER_CACHE_VPS_LIVENESS_URL"]
+            try:
+                sitedata = urllib3.request("GET", vps_server).json()
```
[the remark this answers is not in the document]
ponyisi: This statement is factually wrong, Copilot

servicex_app/servicex_app/transformer_manager.py
```diff
@@ -410,6 +414,31 @@ def persistent_volume_claim_exists(self, claim_name, namespace):
...
+                if thissite not in sitedata:
+                    current_app.logger.error(f"{thissite} is not in VPS liveness data")
+                servers = sorted(
+                    [_["address"] for _ in sitedata[thissite].values() if _["live"]]
+                )
+                current_app.logger.info(f"Live xCache servers are {servers}")
+                current_app.config["TRANSFORMER_CACHE_PREFIX"] = ",".join(servers)
+            except Exception as e:
+                current_app.logger.error(f"Exception retrieving site data: {e}")
+
```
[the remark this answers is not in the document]
ponyisi: No reason to do this and we do not want to throw an exception into the the transformation submission because this update failed for whatever reason

docs/deployment/reference.md
```diff
@@ -93,7 +93,10 @@ parameters for the [rabbitMQ](https://github.com/bitnami/charts/tree/master/bitn
...
 | `minio.apiIngress.hostname`                   | Hostname associate with ingress controller                                                                                                                                                                       | nil                                            |
-| `transformer.cachePrefix`                     | Prefix string to stick in front of file paths. Useful for XCache                                                                                                                                                 |                                                |
+| `transformer.cachePrefix`                     | Prefix string to stick in front of file paths. Useful for XCache                                                                                                                                                 | nil                                               |
+| `transformer.cacheVPSSite`                     | Specify a Virtual Placement Service site whose XCaches we should use. Will update automatically if list changes                                      | nil                                               |
```
gordonwatts: Add documentation somewhere to explain how `cachePrefix` and `cacheVSxxx` interact.

docs/deployment/reference.md
```diff
@@ -93,7 +93,10 @@ parameters for the [rabbitMQ](https://github.com/bitnami/charts/tree/master/bitn
...
 | `minio.apiIngress.hostname`                   | Hostname associate with ingress controller                                                                                                                                                                       | nil                                            |
-| `transformer.cachePrefix`                     | Prefix string to stick in front of file paths. Useful for XCache                                                                                                                                                 |                                                |
+| `transformer.cachePrefix`                     | Prefix string to stick in front of file paths. Useful for XCache                                                                                                                                                 | nil                                               |
```
gordonwatts: You use "nil" here, but "null" in the code itself. Should we keep consistent?

gordonwatts (changes requested):
The only thing that wasn't really obvious to me was how the two cache schemes would interact (prefix vs VS). Other than that, this should go in!

docs/deployment/reference.md
ponyisi: This is consistent within the documentation (you can see the previous line) - should get fixed probably but this should be a global fix

gordonwatts (commented):
I like this - but I had only one question - we have things like cachePrefix, etc. - lots of ways to modify the cache. It looks like they could step on each others feet. We should have some brief documentation in markdown on how it all works together. See my unresoved comment in the previous review.

ponyisi:
The documentation now does indicate which option takes priority. Is there somewhere else you would like this explained?

> I like this - but I had only one question - we have things like cachePrefix, etc. - lots of ways to modify the cache. It looks like they could step on each others feet. We should have some brief documentation in markdown on how it all works together. See my unresoved comment in the previous review.

gordonwatts (approved):
Finally saw the change you added! Sorry! And this is great, thanks!!

PR #1134, opened 2025-08-04
Add models and routes for reporting dataset lookup errors

ponyisi:
Companion to https://github.com/ssl-hep/ServiceX_DID_Finder_lib/pull/38 . Adds a new dataset lookup failure endpoint and updates models as necessary. Still missing an endpoint to query what the error was (for client information) and tests.

### DIFF
--- a/servicex_app/servicex_app/models.py
+++ b/servicex_app/servicex_app/models.py
@@ -157,6 +157,7 @@ class TransformStatus(Enum):
     complete = ("Complete", True)
     fatal = ("Fatal", True)
     canceled = ("Canceled", True)
+    bad_dataset = ("Bad Dataset", True)
 
     def __init__(self, string_name, is_complete):
         self.string_name = string_name
@@ -404,6 +405,9 @@ class DatasetStatus(str, Enum):
     created = "created"
     looking = "looking"
     complete = "complete"
+    does_not_exist = "does_not_exist"
+    bad_name = "bad_name"
+    internal_failure = "internal_failure"
 
 
 class Dataset(db.Model):
--- a/servicex_app/servicex_app/resources/internal/fileset_error.py
+++ b/servicex_app/servicex_app/resources/internal/fileset_error.py
@@ -1,0 +1,118 @@
+# Copyright (c) 2025, IRIS-HEP
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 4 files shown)


ponyisi:
Actually we already have the endpoint for querying dataset status so all that's needed is the tests

servicex_app/servicex_app/models.py
```diff
@@ -157,6 +157,7 @@ class TransformStatus(Enum):
     complete = ("Complete", True)
     fatal = ("Fatal", True)
     canceled = ("Canceled", True)
+    bad_dataset = ("Bad Dataset", True)
```
gordonwatts, line 160: Maybe I'm just getting going in the review here, but shouldn't `empty` be here?

gordonwatts (changes requested):
But wait - where are the tests?

servicex_app/servicex_app/models.py
ponyisi: ? I don't think "existing but empty" dataset should be an error condition?

gordonwatts: This is a very common case where the dataset has had all its files deleted. Their isn't an error in the sense this is by design in rucio. But it is very much not what the user wants or intends to run on. Asking servicex to run on zero files seems... like it should be an error.

ponyisi: @gordonwatts I'm not really comfortable imposing this kind of logic server-side, as it's a statement about ATLAS semantics... I'd say a more insidious issue is dataset _partially_ deleted and we have no checks for that. On the whole we have no idea how many files "should" be there if they're not registered any more. I would be OK with the client and the accessory code putting up a WARNING but on the server side I don't think this should be treated as an error. What I would like to ensure is that we distinguish "dataset doesn't even exist" or "Rucio servers are broken"; "the request is valid but returns no files" would then be the remaining case in which "status complete but zero files" would apply.

gordonwatts: I'm ok with this approach. If only to get this PR moving again. :-)

I still think anyone doing physics analysis will want to know, in an in-your-face way, if a dataset has zero files. Likely they would want to also know if it was partial, but we have no way of detecting that currently. Both of these conditions are usually considered very bad for the health of your results.

ponyisi:
Complete, Gordon signed off offline

PR #1140, opened 2025-08-15
Switch back to python image since trixie base is now available

ponyisi:
Avoid rebuilding python parts of base image for xrootd did container


gordonwatts (approved):
Looks good!

PR #1163, opened 2025-09-23
support custom science images

### LINKED ISSUE #1162: Support custom TopCP images
Backend work for: https://github.com/ssl-hep/ServiceX_frontend/issues/658

MattShirley:
Backend work for supporting custom TopCP images from whitelisted Docker registry repositories: https://github.com/ssl-hep/ServiceX_frontend/issues/658

Allowed images can be defined using the `app.allowedDockerRegistries` value, where each key should be a Docker registry and each value should be a list of repositories/images to validate against. Each string in the list will be tested to see if it is a valid prefix of the custom image. If it is, it will be allowed. The default value will allow any tags for the Docker Hub `sslhep/` repository.

```yaml
app:
  allowedImagePrefixes:
    - "sslhep/"
    - "docker.io/sslhep"
```

Sample files that work with a custom TopCP image but not with the default/base image:

```yaml
# reco.yaml
CommonServices:
  runSystematics: False

PileupReweighting: {}

EventCleaning: {}

Trigger:
    triggerChainsPerYear:
        '2015':
            - 'HLT_2mu6_bBmumuxv2'
            - 'HLT_2mu4_bBmumuxv2'
        '2016':
            - 'HLT_2mu6_bBmumux_BsmumuPhi_delayed'
            - 'HLT_2mu6_bBmumux_BsmumuPhi_delayed_L1BPH-2M9-2MU6_BPH-2DR15-2MU6'
            - 'HLT_2mu6_bBmumux_BsmumuPhi_L1BPH-2M9-2MU6_BPH-2DR15-2MU6'
            - 'HLT_mu10_mu6_bBmumux_BsmumuPhi_delayed'
            - 'HLT_2mu6_bBmumuxv2'
            - 'HLT_2mu6_bBmumuxv2_delayed'
        '2017':
            - 'HLT_mu6_mu4_bBmumux_BsmumuPhi_L1BPH-2M9-MU6MU4_BPH-0DR15-MU6MU4'
            - 'HLT_2mu6_bBmumux_BsmumuPhi_L1BPH-2M9-2MU6_BPH-2DR15-2MU6'
            - 'HLT_2mu6_bJpsimumu_L1BPH-2M9-2MU6_BPH-2DR15-2MU6'
            - 'HLT_2mu10_bBmumuxv2'
        '2018':
            - 'HLT_mu6_mu4_bBmumux_BsmumuPhi_L1BPH-2M9-MU6MU4_BPH-0DR15-MU6MU4'
            - 'HLT_2mu6_bBmumux_BsmumuPhi_L1BPH-2M9-2MU6_BPH-2DR15-2MU6'
            - 'HLT_2mu6_bJpsimumu_L1BPH-2M9-2MU6_BPH-2DR15-2MU6'
            - 'HLT_2mu10_bBmumuxv2'
    noFilter: True
    noGlobalTriggerEff: True

Muons:
  - containerName: AnaMuons
    WorkingPoint:
      - selectionName: loose
        quality: Loose
        isolation: NonIso
      - selectionName: medium
        quality: Medium
        isolation: NonIso

BPhyVertex:
  - vertexName: 'BPHY28BsKKMuMuCandidates'
    decorations: []


BPHY28:
  - vertexName: 'BPHY28BsKKMuMuCandidates'
    inputMuons: 'AnaMuons'
    decorateTrackParameters: True
    decoratePixelHits: True
    decorateDedx: True

# After configuring each container, many variables will be saved automatically.
Output:
  treeName: 'reco'
  vars: []
  metVars: []
  containers:
      # Format should follow: '<suffix>:<output container>'
#      vtx_: 'BPHY28BsKKMuMuCandidates'
      mu_: 'AnaMuons'
      '': 'EventInfo'
  commands: []

AddConfigBlocks:
  - modulePath: 'TopCPToolkit.BPHY28Config'
    functionName: 'BPHY28Config'
    algName: 'BPHY28'
    pos: 'Output'
  - modulePath: 'TopCPToolkit.BPhyVertexConfig'
    functionName: 'BPhyVertexConfig'
    algName: 'BPhyVertex'
    pos: 'Output'
```

```yaml
# spec.yaml
General:
  OutFilesetName: "fileset_ntuple_v1_af"
  OutputDirectory: "."

Sample:
  ## data18
  - Name: data18_periodD
    Dataset: !Rucio data18_13TeV:DAOD_BPHY28.44180527._000001.pool.root.1
    Query: !TopCP |
      reco=reco.yaml,
      image=sslhep/servicex_science_image_topcp:25.2.50-v2.18.0_v0.2

```


ArturU043:
Hi, 

I have a comment about the `spec.yaml` variable used for image selection. 
```yaml
docker_image=sslhep/servicex_science_image_topcp:25.2.50-v2.18.0_v0.2
```
Shouldn't the user specify both `AB` and `TopCP` versions separately? 
And `docker_image` would parse them with this pattern to fetch the actual image.
This may require a large variety of science images on Docker Hub... 

Or at least there should be a way for the user to effortlessly list all available images on the client side.

MattShirley:
@ArturU043 I don't think there's any plan to change the science image build process. Do you anticipate needing to build the same custom images for different AB/TopCP combinations?

As far as adding a command to list available images, I can create a ticket for that.

BenGalewsky:
Please update [docs/deployment/reference.md](https://github.com/ssl-hep/ServiceX/blob/develop/docs/deployment/reference.md)

BenGalewsky:
Is there any way we can move the image selection retrieval and verification into the TopCP CodeGen? I really don't like having a hardcoded test for `topcp` in the submit transform handler.

I don't think we need to change the app at all to support this special case

servicex_app/servicex_app/code_gen_adapter.py
```diff
@@ -43,7 +45,11 @@ def post_request(self, post_url, post_obj):
         return result
 
     def generate_code_for_selection(
-        self, request_record: TransformRequest, namespace: str, user_codegen_name: str
+        self,
+        request_record: TransformRequest,
+        namespace: str,
+        user_codegen_name: str,
+        custom_image: Optional[str] = None,
```
BenGalewsky: Why is`custom_image` still here?

helm/servicex/templates/app/configmap.yaml
```diff
@@ -183,6 +183,11 @@ data:
     CODE_GEN_IMAGES = { {{ join "," $code_gen_images }} }
     {{- end }}
 
+    {{- if .Values.codeGen.topcp.enabled }}
+    # TopCP custom image configuration
+    TOPCP_ALLOWED_IMAGES = {{ .Values.codeGen.topcp.allowedImages | toJson | quote }}
```
BenGalewsky: Do we still need this in the App config?

code_generator_TopCPToolkit/servicex/TopCP_code_generator/request_translator.py
```diff
@@ -26,14 +26,34 @@
...
+
 from . import query_translate
 from servicex_codegen.code_generator import (
     CodeGenerator,
     GeneratedFileResult,
     GenerateCodeException,
 )
 
+def validate_custom_docker_image(image_name: str) -> bool:
+    allowed_images_json = os.environ.get("TOPCP_ALLOWED_IMAGES")
```
BenGalewsky: Could this be moved to the codegen's `__init__` so that if there is a formatting error in the ALLOWED_IMAGES it will fail at startup instead of the first time someone submits?

MattShirley: The function call can't go in `__init__.py` because there's nothing to compare it to. However, I added a check in `create_app` to validate the JSON parsing of of TOPCP_ALLOWED_IMAGES. And I removed type guarding from `validate_custom_docker_image`.

ponyisi:
So I have to admit I had imagined a more generic framework (i.e. instead of just enabling checks for TopCP, have the `servicex_app` validate image requests across all possible codegens). I guess maybe this is something that we could refactor later.

MattShirley:
@ponyisi could you elaborate on exactly what you think should be extracted from the codegen?

It'd be nice if this entire blob could get moved into `submit.py`, however the codegen is supposed to set the image. Would simply ignoring the codegen's transformer image be acceptable across the board if `image` and `registry` are set?
```python3
        if "image" in jquery:
            image = jquery["image"]
            validate_custom_docker_image(image)

            if "registry" in jquery:
                image = f"{jquery['registry']}/{image}"

            results.image = image
```

ponyisi:
@MattShirley not really sure I understand... we already have a hook ( https://github.com/ssl-hep/ServiceX/blob/develop/servicex_app/servicex_app/resources/transformation/submit.py#L229 ) where the image requested by any codegen can potentially be validated. At this point it's a common code path regardless of whether the codegen is using the default image or is overriding. If it's an unacceptable image, we should just terminate the request. One can just whitelist the standard science images by default so there's no reason to make a distinction between default and override.

MattShirley:
@ponyisi `check_image_exists` currently only validates that the image actually exists. And there are pretty big limitations to it, because there is no standardized way to verify an image exists given just an image name (including the registry) over HTTP. So as part of this PR, to support our GitLab docker registry, I had to code a custom HTTP lookup.

This block could be inserted right before the `TRANSFORMER_VALIDATE_DOCKER_IMAGE` check, and then any codegen would be able to support custom images (this is my interpretation of your request to centralize the code inside `servicex_app`):
```python3
        if "image" in jquery:
            image = jquery["image"]
            validate_custom_docker_image(image)

            if "registry" in jquery:
                image = f"{jquery['registry']}/{image}"

            results.image = image
```

I think this `validate_custom_docker_image` check should only be applied when the custom fields are set and shouldn't be merged with the other validation logic. This is because the `TRANSFORMER_VALIDATE_DOCKER_IMAGE` check is optional. It's only use is to catch failures before they occur. `validate_custom_docker_image` in contrast actually enforces the config's business logic regarding which registries and image prefixes are allowed.

ponyisi:
@MattShirley I didn't mean to merge the two logics together, I was just pointing out that we already have a location in the code where the information is available to do the check. So inserting it just before the `TRANSFORMER_VALIDATE_DOCKER_IMAGE` check sounds fine.

I disagree that the check should only be applied when custom fields are set; this breaks the principle that the `servicex_app` should not know anything about the query format for the codegens. It's more complicated to figure out if the requested image is non-default than to just check all image requests.

MattShirley:
@ponyisi this PR has been updated to support CERN's GitLab instance. It also now validates all science images.

servicex_app/servicex_app/docker_repo_adapter.py
```diff
@@ -26,41 +26,101 @@
...
-        query = f"{self.registry_endpoint}/v2/repositories/{repo}/{image}/tags/{tag}"
-        r = requests.get(query, timeout=REQUEST_TIMEOUT)
-        return r
+    def get_image_by_tag(
+        self, repo: str, image: str, tag: str, registry: str
+    ) -> requests.Response:
+        if registry == "docker.io":
+            return self._dockerhub_get_image_by_tag(repo, image, tag)
+
+        if registry == "gitlab-registry.cern.ch":
```
BenGalewsky: Is there really no way to make this generic? This is pretty 🤢

MattShirley: I added `crane` to the image and simplified this considerably

helm/servicex/values.yaml
```diff
@@ -33,6 +33,13 @@ app:
   tokenExpires: false
   validateTransformerImage: true
   sqlalchemyEngineOptions: null
+  allowedDockerRegistries:
+    "docker.io":
+      allowedImagePrefixes:
```
ponyisi: is a prefix mandatory or can this just be "" ?

servicex_app/servicex_app/docker_repo_adapter.py
```diff
@@ -25,42 +25,30 @@
...
-    def __init__(self, registry_endpoint="https://hub.docker.com"):
-        self.registry_endpoint = registry_endpoint
-
-    @servicex_retry()
-    def get_image_by_tag(self, repo: str, image: str, tag: str) -> requests.Response:
-        query = f"{self.registry_endpoint}/v2/repositories/{repo}/{image}/tags/{tag}"
-        r = requests.get(query, timeout=REQUEST_TIMEOUT)
-        return r
-
     def check_image_exists(self, tagged_image: str) -> bool:
```
ponyisi: would it be worth memo-ifying this?

servicex_app/servicex_app/resources/transformation/submit.py
```diff
@@ -223,7 +253,24 @@ def post(self):
                 request_rec, namespace, user_codegen_name
             )
 
-            request_rec.image = codegen_transformer_image
+            try:
+                jquery = json.loads(args["selection"])
```
ponyisi: This is not generically going to work for all codegens, many do not use JSON for their selection and even when they are they do not all have an "image" key.

servicex_app/servicex_app/resources/transformation/submit.py
```diff
@@ -223,7 +253,24 @@ def post(self):
...
+            except json.decoder.JSONDecodeError:
+                jquery = {}
+
+            registry = jquery.get("registry", "docker.io")
+
+            transformer_image = codegen_transformer_image
+            if "image" in jquery:
```
ponyisi: Again I don't really understand the point of doing this. If the codegen is doing the correct thing, `codegen_transformer_image`, which is under the control of the codegen which is interpreting the selection request, will have the image we want. We should not have any logic here that tries to peek into the selection, this must be kept opaque at this level. Basically, we should take `codegen_transformer_image` and prefix it with `docker.io` if necessary, that's it.

code_generator_TopCPToolkit/servicex/TopCP_code_generator/request_translator.py
```diff
@@ -62,11 +64,15 @@ def generate_code(self, query, cache_path: str):
...
+        jquery = json.loads(query)
+        query_translate.generate_files_from_query(jquery, query_file_path)
+
         shutil.copyfile(
             capabilities_path,
             os.path.join(query_file_path, "transformer_capabilities.json"),
         )
 
-        query_translate.generate_files_from_query(query, query_file_path)
+        results = GeneratedFileResult(_hash, query_file_path)
```
ponyisi, line 76: the `GeneratedFileResult` constructor should take an additional `image=jquery.get("image", None)` argument

helm/servicex/values.yaml
MattShirley: Right now a prefix is mandatory however it can just be `""`. I don't think we should include anything with `""` in the default `values.yaml` but a cluster operator could choose to do so. Do you think we should keep KyungEon's GitLab repository by default? @kyungeonchoi do you know how easy it would be to setup a shared namespace in CERN's GitLab like `sslhep-customimages` that isn't tied to a specific user name?

code_generator_TopCPToolkit/servicex/TopCP_code_generator/request_translator.py
MattShirley: The image handling logic has been added below this line now.

servicex_app/servicex_app/resources/transformation/submit.py
MattShirley: This has been moved to the TopCP Code Gen's standard image setting logic.

servicex_app/servicex_app/resources/transformation/submit.py
MattShirley: This has been moved to the TopCP Code Gen's standard image setting logic. Now the code moves immediately to validate the image name.

servicex_app/servicex_app/docker_repo_adapter.py
MattShirley: Good catch. Not only will that give us performance upgrade, but it will give us significant protection from temporary networking outages.

code_generator_TopCPToolkit/servicex/TopCP_code_generator/request_translator.py
```diff
@@ -62,11 +64,21 @@ def generate_code(self, query, cache_path: str):
...
+
+        if jquery.get("image"):
```
ponyisi, line 78: Again, just really dumb question, why do we need separate `registry` and `image` arguments if we just squish them? (I'm being kind of insistent on this point because the "standard" syntax for users to refer to images - e.g. in our own helm charts - just gives a single string with an optional registry in front, it doesn't explicitly separate registry and image name, and I don't see a reason to introduce new syntax of our own here?)

helm/servicex/values.yaml
ponyisi: Personally I'd tend to say we just whitelist the full `gitlab-registry.cern.ch` by default

code_generator_TopCPToolkit/servicex/TopCP_code_generator/request_translator.py
MattShirley: The registry logic has been completely stripped out of this PR and is now assumed to be part of the image value.

ponyisi (approved):
OK, let's go

Issue #927, opened 2024-11-24
Distinguish the cases of "zero files in dataset", "number of files is unknown", and "dataset does not exist"


ponyisi:
There seem to be multiple places in the user interface (e.g. the dashboard) where a dataset having zero files (e.g. when a Rucio dataset does not exist) is reported as "Unknown". 
![image](https://github.com/user-attachments/assets/a96f0477-a8ff-4e46-8b54-5e5fb82f1fa7)
In the case of Rucio DIDs this is often because the dataset does not exist, which we should handle in some more obvious way than reporting an "Unknown" number of files. In some cases the Rucio dataset might exist but be empty, in which case we should properly report zero files.

ponyisi:
Also add the case "DID finder crashed"...

ponyisi:
Proposal:

Database models:
* the `DatasetStatus` enum gains the extra value `lookup_failed`
* the `TransformStatus` enum gains the extra value `bad_dataset`; this will be considered to be a termination status for the transformation. (It will require client support ...)

We add another `status` field to the JSON sent to `FilesetComplete`. If the lookup is a success the field will say so. If the DID finder decides that the input DID specification is invalid, this field will mark that, and if the DID finder crashed in the lookup, there will be a third marker value.

Based on this:
* if the DID finder decides the DID is bad it will raise an exception (specified in the DID finder library) which the DID finder library will catch and use to return the "invalid DID" status. The ServiceX app will then mark the dataset with `lookup_failed`, and all referring transforms with `bad_dataset`. The dataset will be marked as stale (so in case it _does_ exist in the future we will be able to find it). 
* if the DID finder fails otherwise the library will return the "lookup failure" status, at which point the dataset will be marked as `lookup_failed` and stale, and referring transforms will be marked as "fatal" completion.

gordonwatts:
See https://github.com/ssl-hep/ServiceX/pull/1134 for the infrastructure - partially completes this.

ponyisi:
Addressed by #1173 and https://github.com/ssl-hep/ServiceX_frontend/pull/674

PR #1173, opened 2025-10-07
Use standard DID exceptions in finders

ponyisi:
The new version of the DID finder library has standard exceptions that are used to signal specific error conditions. This will allow us to fail requests when they make bad requests, see #927 .


gordonwatts (approved):
This looked good to me - can't wait for it to go in!

PR #1187, opened 2025-10-17
Migrate DB enum to support bad_dataset

ponyisi:
Add missing DB enum support for the bad dataset status. With this things work


BenGalewsky (approved):
Ah, thanks

PR #1190, opened 2025-10-23
set RABBIT_MQ_URL value in helm

MattShirley:
My transformers were hanging due to issues establishing a connection to RabbitMQ. It came from the recent changes to the boot.sh script where `$RABBIT_MQ_URL` is assumed to exist and contain the RabbitMQ connection string. However, this value is only present when secrets are configured.

The change here sets a RABBIT_MQ_URL value in the servicex app deployment even if secrets aren't available.


ponyisi (approved):
Thanks for catching it and for the fix

PR #1191, opened 2025-10-27
Update coverage configuration

ponyisi:
Enable uploading of coverage data to codecov.io


ponyisi:
For some reason the `x509_secrets` container has no test, but otherwise it looks like we have everything in with this.

gordonwatts (approved):
The real test is if this works to generate coverage. To my eye, this looks great - way more sophisticated than what I normally create for coverage.

PR #1192, opened 2025-10-28
Fix RabbitMQ secrets password on 1.7.3

ponyisi:
Overlooked an if/else for rabbitmq password


gordonwatts (approved):
Ah - got it - secrets are being distributed differently in prod.

PR #1201, opened 2025-11-04
Have sideband tasks in app container attempt indefinite reconnection to RabbitMQ

ponyisi:
If the RabbitMQ container dies for some reason, by default celery will retry connections 100 times then stop. This means a sufficiently long outage will leave the `app` in a state where jobs can be submitted, but the file list never gets dispatched to the transformers because that worker has exited. 

The update here is only for the worker spun up in the `app` container. The DID finders in principle have the same problem but since the Celery workers are the primary tasks in those containers, when they exit k8s will restart them so we don't see the issue there. I guess something similar happens with the transformer sidecars although that is certainly going to be ugly.

### DIFF
--- a/servicex_app/boot.sh
+++ b/servicex_app/boot.sh
@@ -21,10 +21,12 @@ else
 fi
 [ -d "/default_users" ] && python3 servicex/cli/create_default_users.py
 
-celery --config servicex_app.celery.celeryconfig \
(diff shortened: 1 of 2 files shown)


BenGalewsky (approved):
Looks like a good solutuon

Issue #1166, opened 2025-10-06
Cleanup Service Should Keep Us Below a Storage Threshold


BenGalewsky:
_As a site admin I want the cleanup service to delete oldest results first, but keep going to meet a storage threshold so I can insure I never run out of space_

Extend the current cleanup service to use space as an additional criteria

PR #1202, opened 2025-11-07
1166 cleanup space limit

BenGalewsky:
# Problem
The current data lifecycle workflow just uses the cutoff date to clean up old transforms. If a site is running low on space, then we need to take more aggressive action and keep deleting the oldest transforms until we are back below an admin specified threshold. 

# Description
Solves #1166

1. Add a new setting in the app config; `MAX_DESIRED_TRANSFORM_CACHE_SIZE` - this is a string and can have units such as Mb, Gb, or Tb
2. Added a new database method to the TransformRequest class to report the total size of the server-side cache
3. Added a new query to the TransformRequest class to find the oldest transform request where the accumulated sizes first gets us a certain number of bytes.
4. Updated the `post` method in the DataLifecycle resource to:
   a. See how many bytes are currently in the cache
   b. Determine how many bytes we need to delete to get us below the threshold
   c. Find the cutoff time that would get us there
   d. Compute the cutoff date that is the earliest of cutoff date and the date to get us below threshold
   e. Provide more complete report back to the calling service of how many bytes we cleared, etc.
5. Updated helm chart to add this setting to the config file

The App's log statements didn't include the timestamp. Updated the logger to make sure our logs can help us debug any issues with this workflow

## How to Test
I created a custom `values.yaml` file that exercises this workflow pretty easily:
```yaml
dataLifecycle:
    enabled: true
    schedule: "*/2 * * * *"
    retention: "3 minutes ago"
    maxDesiredCacheSize: "10Mb"
```

You can see jobs being created and view their logs. I saw that it handles and empty cache well. Ran a couple of transforms and see that it eventually deletes the oldest transforms and reports new cache size


helm/servicex/values.yaml
```diff
@@ -261,3 +261,7 @@ dataLifecycle:
   # See https://www.geeksforgeeks.org/date-command-linux-examples/#4-how-to-display-past-dates
   # for examples
   retention: "7 days ago"
+
+  # The cleanup service will go beyond the retention date to keep the server side cache size below
+  # this threshold. Specify this as a string with Mb, Gb, or Tb units in the string
+  maxDesiredCacheSize: "100Gb"
```
ponyisi: This is rather small for a real deployment, no? Even our testing instances have O(400 GB) for the cache ...

servicex_app/servicex_app/resources/internal/data_lifecycle_ops.py
```diff
@@ -101,13 +101,65 @@ def delete_orphaned_datasets(self, session: Session):
 
         return deleted_datasets
 
+    @property
+    def max_desired_cache_size(self):
+        value = current_app.config.get("MAX_DESIRED_TRANSFORM_CACHE_SIZE")
+        if not value:
+            return None
+        value = value.strip().lower()
+        if value.endswith("tb"):
```
ponyisi: Perhaps PB should be an option?

ponyisi:
Very naively - where does the cron job actually get scheduled?

helm/servicex/values.yaml
BenGalewsky: I would assume that there is no universal default and site admins will need to set this. I can update this to make 1Tb the default?

BenGalewsky:
> Very naively - where does the cron job actually get scheduled?

It's in the [helm chart](https://github.com/ssl-hep/ServiceX/blob/1166_cleanup_space_limit/helm/servicex/templates/data-lifecycle/cronjob.yaml)

servicex_app/servicex_app/resources/internal/data_lifecycle_ops.py
BenGalewsky: Good idea. Done

helm/servicex/values.yaml
BenGalewsky: Changed to 1Tb

ponyisi:
I have no idea why the `cronjob.yaml` doesn't appear in the "Files changed" tab, but I'll assume it's just a GitHub display glitch then...

(Edit: oh, this lingers from some older iteration of this?)

ponyisi (approved):

PR #1203, opened 2025-11-08
Enable reading ATLAS Opendata via Rucio

ponyisi:
ATLAS Opendata on Rucio was blocked because the endpoint was marked as `type=SPECIAL`. This actually blocks a number of endpoints with unique datasets (mostly at CERN). Remove that restriction for now, since the `rucio` command line would be able to download them. Also improve the logic to catch when replicas are unavailable.


gordonwatts (approved):
I had not appreciated that Special was a designation in rucio rather than by us. We probably should understand a little better what that means, I agree. I will ask Brian, he may know RobG is another good person to ask.

BenGalewsky (approved):
I don't understand the full context of this change, but it certainly does what is claimed

Issue #1174, opened 2025-10-07
Consider webdav replicas


ponyisi:
Currently the rucio DID finder only considers http and xrootd replicas. However the one protocol that all sites must support is webdav. This is not a problem to use in principle since the files can be accessed over http. Add support for these replicas.

PR #1204, opened 2025-11-09
Enable `davs` replicas in the Rucio DID finder

ponyisi:
Accept `davs` replicas and prioritize these first. (To be discussed if we want to keep this order, but it's inspired by discussions at the 2025 retreat.)

### DIFF
--- a/did_finder_rucio/src/rucio_did_finder/rucio_adapter.py
+++ b/did_finder_rucio/src/rucio_did_finder/rucio_adapter.py
@@ -173,7 +173,7 @@ class RucioAdapter:
             try:
                 reps = self.replica_client.list_replicas(
                     [{"scope": ds[0], "name": ds[1]}],
-                    schemes=["root", "http", "https"],
+                    schemes=["davs", "root", "http", "https"],
                     metalink=True,
                     sort="geoip",
                     rse_expression="istape=False\\type=SPECIAL",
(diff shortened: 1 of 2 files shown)


transformer_sidecar/src/transformer_sidecar/transformer.py
```diff
@@ -134,6 +135,9 @@ def transform_file(
     # Prioritize the replicas
     _file_paths = prioritize_replicas(paths)
 
+    # Change davs to https
+    _file_paths = change_davs_to_https(paths)
```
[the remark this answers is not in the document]
MattShirley: Seconding this. It looks like the results of the previous line are lost.

ponyisi:
Addresses #1174

ponyisi:
Following discussion I am going to deprioritize davs vs rootd for now, but keep the replicas in the list.

did_finder_rucio/src/rucio_did_finder/rucio_adapter.py
```diff
@@ -173,7 +173,7 @@ def list_files_for_did(self, did):
             try:
                 reps = self.replica_client.list_replicas(
                     [{"scope": ds[0], "name": ds[1]}],
-                    schemes=["root", "http", "https"],
+                    schemes=["davs", "root", "http", "https"],
```
gordonwatts, line 176: THis looks like right now `davs` is top prioprity, but I think I saw you wanted to switch the order.

gordonwatts (approved):
Approving - though it looks like the order is not what you wanted.

did_finder_rucio/src/rucio_did_finder/rucio_adapter.py
ponyisi: This is not how the priority gets set (that is done in the sidecar code)

Issue #1167, opened 2025-10-06
DID Finder Cache should expire after a week


gordonwatts:
The current DID Finder cache remains forever.

* The locations of files on the GRID periodically move
* Datasets replica numbers change over time as their popularity fluctuates

This causes open file errors (or missed files because new replicas aren't picked up). To counter this, have the DID finder's cache reset periodically.

Start with a week.

ponyisi:
Suggest this is done by adding a Kubernetes cronjob whose schedule is configurable via `values.yaml`. Ilija suggested this run every day by default; I'm not sure what's optimal but I guess we can offload that to the sites.

gordonwatts:
I like the idea of running that once a day. Do we have any way to retire something in the list? It would be bad if it just got longer and longer over time until we were querying every dataset for years. ;-)

ponyisi:
We have an option to only list datasets that aren't stale, so that should be fine.

BenGalewsky:
Closed via #1211

PR #1211, opened 2025-11-17
Dataset cache purge cronjob

ponyisi:
Address #1167 , except the cache lifetime is reduced to one day.


helm/servicex/templates/dataset-lifecycle/cronjob.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: batch/v1
```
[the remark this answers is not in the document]
ponyisi: Should always be enabled IMO

servicex_app/servicex_app/resources/internal/dataset_lifecycle_ops.py
```diff
@@ -0,0 +1,59 @@
...
+from ..datasets.get_all import get_all_datasets
+from ..datasets.delete_dataset import delete_dataset
+
+
+class DatasetLifecycleOps(ServiceXResource):
+    def post(self):
+        """
+        Obsolete cached datasets older than N hours
+        """
+        now = datetime.now()
```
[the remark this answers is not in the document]
ponyisi: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/ssl-hep/ServiceX/pull/1211#discussion_r2535325824)

servicex_app/servicex_app/resources/datasets/delete_dataset.py
```diff
@@ -30,18 +30,22 @@
 from servicex_app.resources.servicex_resource import ServiceXResource
 
 
-class DeleteDataset(ServiceXResource):
-    @auth_required
-    def delete(self, dataset_id):
-        dataset = Dataset.find_by_id(dataset_id)
+def delete_dataset(dataset_id):
+    dataset = Dataset.find_by_id(dataset_id)
```
BenGalewsky: Move this to models.py since this now needs to be reused between two resources. We shouldn't be calling methods in one resource from another

servicex_app/servicex_app/resources/internal/dataset_lifecycle_ops.py
```diff
@@ -0,0 +1,61 @@
...
+        Obsolete cached datasets older than N hours
+        """
+        now = datetime.now(timezone.utc)
+        try:
+            age = float(request.get_json().get("age", 24))
+        except Exception:
+            return {"message": "Invalid age parameter"}, 422
+        delta = timedelta(hours=age)
+        datasets = (
+            get_all_datasets()
```
BenGalewsky: Why can't this be `Dataset.get_all()` from models.py? We shouldn't be calling methods in one resource from another resource

docs/deployment/reference.md
```diff
@@ -111,9 +111,14 @@ parameters for the [rabbitMQ](https://github.com/bitnami/charts/tree/master/bitn
...
 | `dataLifecycle.retention`                    | We will archive any transforms older than this. Use the gnu date command --date argument. See [date command](https://www.geeksforgeeks.org/date-command-linux-examples/#4-how-to-display-past-dates) for examples. | 7 days ago                                     |
 | `dataLifecycle.maxDesiredCacheSize`          | If the server-side cache is larger than this cleanup service will keep going forward in time to delete transforms. Specify units as Mb Gb, Tb or Pb                                                                | "1Tb"                                          |
+| `datasetLifecycle.image`                         | Default image for dataset cache lifecycle job                                                                                                                                                                  | `curlimages/curl`                              |
```
BenGalewsky: I can't imagine a case where we would want a different image for initiating the dataset clean up as we use for the data lifecycle job.
In fact, I had imagined dataset cleanup to be a subset of data lifecycle, so ideally the schedule should just be a subitem of `dataLifecycle`

So, maybe 
```
dataLifecycle:
      image: curlimages/curl
      tag: "3.10"
      pullPolicy: Always

    transformCleanup:
        enabled: true
         schedule: "* */8 * * *"
         retention: "7 days ago"
         maxDesiredCacheSize: "1Tb"
    datasetCleanup:
        enabled: true
        schedule: "0 * * * *"
        cacheLifetime: 24
```

BenGalewsky (changes requested):
Ok - glad you were able to tackle this. 
Just a couple of issues:
1. Don't call methods in one resource from another. Migrate common database operations to `models.py`
2. Make Transform cleanup and dataset cleanup siblings in the helm values `dataLifecycle` properties

docs/deployment/reference.md
ponyisi: The data lifecycle and dataset cache tasks are conceptually quite separate, one is handling ServiceX outputs and one is handling input replicas and there's no real reason their schedules should be at all related. In fact you might want to disable the data lifecycle and keep the dataset cache cleaning. (I had considered doing a single task but unless we make it explicitly a single "general" job that just does all periodic cleanups I think this is adding unhelpful complication.)

ponyisi: The data lifecycle and dataset cache tasks are conceptually quite separate, one is handling ServiceX outputs and one is handling input replicas and there's no real reason their schedules should be at all related. In fact you might want to disable the data lifecycle and keep the dataset cache cleaning. (I had considered doing a single task but unless we make it explicitly a single "general" job that just does all periodic cleanups I think this is adding unhelpful complication.)

BenGalewsky: I agree they both need their own schedules and could be enabled separately - all I'm saying is that they don't need different docker images, so if we collect these two operations in the values.yaml as nested under a common root level property that way they can share the docker image specification. My formatting in the above comment got messed up. I'll edit it

servicex_app/servicex_app/resources/internal/dataset_lifecycle_ops.py
ponyisi: We only call one codepath so OK

BenGalewsky (approved):
Thanks, this looks great!

PR #1212, opened 2025-11-17
fix x509 cronjob

MattShirley:
Fix formatting of x509 CronJob deployment.

Fixes this error I encountered locally in `develop`:

```
Error: INSTALLATION FAILED: 1 error occurred:
	* CronJob.batch "x509-secrets" is invalid: [spec.jobTemplate.spec.template.spec.containers: Required value, spec.jobTemplate.spec.template.spec.restartPolicy: Required value: valid values: "OnFailure", "Never"]
```


ponyisi:
I've requested a second plain `Job` to make sure that the X509 secrets are obtained at the initialization of the ServiceX deployment, not just when the first cron job schedule occurs.

MattShirley:
@ponyisi I added the additional k8s Job.

ponyisi (approved):

PR #1213, opened 2025-11-17
align JWT with ENABLE_AUTH

MattShirley:
The existing setup has no connection between JWT and ENABLE_AUTH.

This PR introduces a `jwt_required_if_auth_enabled` decorator that requires a JWT only if `ENABLE_AUTH=True`. Otherwise, the endpoint will allow traffic without checking for a JWT.


ponyisi:
Hi @MattShirley - I don't think this is quite what I was after, although I guess it addresses another failure mode.

The place where we are currently failing wrt Nebraska is the `jwt_required` decorator itself. _Even if the `optional` argument is specified_, if a token is present in the request, it will attempt to validate it. If `ENABLE_AUTH` is false, this validation will fail and the request will be rejected. The `optional` argument only means that the requests with no token will be accepted. So we need a new decorator that wraps `jwt_required` which skips all the logic if `ENABLE_AUTH` is false (or otherwise perhaps a super-decorator that returns `jwt_required` or a dummy depending on the configuration).

MattShirley:
@ponyisi I added a no-op/dummy pass through decorator

servicex_app/servicex_app/resources/servicex_resource.py
```diff
@@ -58,6 +60,10 @@ def get_requesting_user() -> Optional[UserModel]:
         user = None
         if current_app.config.get("ENABLE_AUTH"):
             user = UserModel.find_by_email(get_jwt_identity())
+            # since jwt is optional, attach user only when jwt available
```
ponyisi: I think the "optional"-ness of the JWT came from the idea that it would handle the case we're now protecting. If `ENABLE_AUTH` is set now, I don't think there is a case where we should not expect a JWT? (In which case we should get rid of the `optional=True` in all cases and this additional protection.)

MattShirley: I have removed both calls with `optional=True`

ponyisi (approved):
Let's go

PR #1215, opened 2025-11-17
Fix logout on OIDC

ponyisi:
This is a mostly cosmetic update: Globus auth doesn't use the correct OIDC logout parameters so Keycloak etc. would complain about a bad redirection URI. This is now fixed.


BenGalewsky (approved):
Ok, thanks <logout>

PR #1218, opened 2025-11-19
Restore OKD-needed permissions for home directories

holzman:
The switch from the base image from bookworm to trixie changed the default umask from 022 to 077. This restores the previous behavior needed for OKD, which runs the container as an ephemeral user with GID 0.


ponyisi (approved):
Thanks. Let's see if integration throws up any issues.

PR #1219, opened 2025-11-22
Update internal state and dependencies

gordonwatts:
- Reset the internal state before executing a run.
- Update the `func-adl-xAOD` dependency version to `^2.2.5`, which has better internal state management.


code_generator_funcadl_xAOD/pyproject.toml
```diff
@@ -8,7 +8,7 @@ packages = [{ include = "code_generator_funcadl_xaod" }]
 
 [tool.poetry.dependencies]
 python = "~3.10"
-func-adl-xAOD = "^2.2.0"
+func-adl-xAOD = "~2.2"
```
ponyisi, line 11: Is this actually better? The original will allow any 2.x, the new one will only show 2.2.x

ponyisi: Ah, no, in this case the caret also requires 2.2.x, so they should be identical. In any case, do you want to keep 2.2.x or allow 2.x ?

gordonwatts: I ended up like that because I had a huge fight with getting the `lock` file to recognize `2.2.5` - caches were not getting updated or something like that. At one point I had it just "2.2.5" and it was still failing.

Tell me what the best practice is for all of ServiceX. It is certainly the case that 2.2.5 is a bug fix release and we really don't want anything less than that running.

But now that the lock file has 2.2.5 in it, I'm not worried about what the `toml` file should look like.

ponyisi: I'm not sure it matters that much - since what matters is the lock file, and we are only going to be updating that under controlled conditions, we *should* know what versions we want to get. So I don't think there's much to say here really, I was just curious what was intended.

ponyisi (approved):

PR #1246, opened 2025-12-19
Slight cleanup in profile creation workflow

ponyisi:
* Make sure that for Globus we make a consistent choice of session and profile email given as there could be multiple emails for a given account
* do not initialize refresh token for CLI-created users in the middle of the profile creation, do it in a sensible place


ponyisi:
Re codecov I don't think there are any tests for the CLI ?

BenGalewsky (approved):
I don't understand the context of this, but I know @ponyisi has been studying this deeply

MattShirley (approved):

PR #1268, opened 2026-01-12
properly quote integers in cronjobs

MattShirley:
This is causing `helm` installation failures due to issues casting integer to string.

```
Error: INSTALLATION FAILED: 1 error occurred:
    * CronJob in version "v1" cannot be handled as a CronJob: json: cannot unmarshal number into Go struct field EnvVar.spec.jobTemplate.spec.template.spec.containers.env.value of type string
```

### DIFF
--- a/helm/servicex/templates/data-lifecycle/cronjob.yaml
+++ b/helm/servicex/templates/data-lifecycle/cronjob.yaml
@@ -19,7 +19,7 @@ spec:
             imagePullPolicy: {{ .Values.scheduledTasks.pullPolicy }}
             env:
               - name: RETENTION_PERIOD
-                value: {{ .Values.scheduledTasks.dataLifecycle.retention }}
+                value: "{{ .Values.scheduledTasks.dataLifecycle.retention }}"
 
             command:
             - /bin/sh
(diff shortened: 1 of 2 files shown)


helm/servicex/templates/data-lifecycle/cronjob.yaml
```diff
@@ -19,7 +19,7 @@ spec:
             imagePullPolicy: {{ .Values.scheduledTasks.pullPolicy }}
             env:
               - name: RETENTION_PERIOD
-                value: {{ .Values.scheduledTasks.dataLifecycle.retention }}
+                value: "{{ .Values.scheduledTasks.dataLifecycle.retention }}"
```
ponyisi: Can we confirm that this works properly if the input is already a quoted string? (double quotes somehow cause weirdness ...)

ponyisi: Like, would it be better to do `{{ .Values.scheduledTasks.datasetLifecycle.cacheLifetime | quote }}` similar to a line near the top of the file?

MattShirley: Good catch. Only the `dataset-lifecycle` cronjob that was causing this error.

MattShirley: I also switched the `dataset-lifecycle` cronjob to use the templating `quote` function

ponyisi (approved):

PR #1280, opened 2026-01-21
Add helm installation name and k8s namespace to local/port-forward.sh

### LINKED ISSUE #1185: local/port-forward.sh should respect custom helm deployment names
# Problem
It is preferred that when you deploy a local helm chart on your laptop that you call it something other than `servicex` since log files all go into the central elasticsearch. I use `servicex-ben` to make it clear.

The port forward script has hardcoded deployment names used in determining the service names. This should be an argument

i.e.

```
        SERVICE="servicex-minio"
```

Should be 
```
        SERVICE="servicex-ben-minio "
```

MattShirley:
Update the `local/port-forward.sh` script (used in local development). To support custom Helm installation names as well as k8s namespaces via the `--helm-name` and `--namespace` arguments respectively.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,6 +1,6 @@
 minikube-mount-app: minikube mount $LOCAL_DIR:/mnt/servicex
 minikube-mount-topcp: minikube mount $LOCAL_DIR/code_generator_TopCPToolkit:/mnt/topcp
-helm-install: sleep 5; cd $CHART_DIR && helm install -f $VALUES_FILE servicex . && tail -f /dev/null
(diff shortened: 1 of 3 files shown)


ponyisi:
This only affects the local test setup configuration, right?

MattShirley:
> This only affects the local test setup configuration, right?

Yes, this script is only used to simplify the Procfile: https://github.com/ssl-hep/ServiceX/blob/develop/Procfile

BenGalewsky (approved):
This still isn't my workflow, but if it works for you, I approve!

PR #1282, opened 2026-01-24
Fix dataset lifecycle jobs

ponyisi:
"Improvements" were made after validation of functioning dataset lifecycle task which in fact broke it - one in the datetime handling and one in the batch job configuration. Fix these (combination is validated to work on testing-3).

Co-Pilot explains what is happening here:
> Note on the change in the PR: The new version adds .replace(tzinfo=None) which strips the timezone information, converting it from a timezone-aware to a timezone-naive datetime. This suggests the code needs to compare against database timestamps that are stored without timezone information.


BenGalewsky:
It would be good to document here what you discovered about timestamps in the database

BenGalewsky (approved):
Thanks -- this looks like an important fact about our database that none of us were aware of!

PR #1283, opened 2026-01-25
Add cleanups for lingering k8s resources

### LINKED ISSUE #1210: Job to kill stuck transformers
If for any reason transformers are stuck for a long time the deployments can hang around forever.
Make a cron job that will periodically shut down very old jobs

ponyisi:
New cronjob to delete transformer deployments, configmaps, and autoscalers older than a configurable age. Will shut down lingering transformer pods as a consequence. 

Resolves #1210


BenGalewsky (approved):
Love it! Thank you

PR #1284, opened 2026-01-25
Fix a bad cron schedule that snuck back in

ponyisi:
Accidentally had a schedule that ran every minute in relevant hours


BenGalewsky (approved):
Ooops, the bug strikes back

PR #1289, opened 2026-01-27
Add resource limits for all objects in helm chart

### LINKED ISSUE #861: Specify resources for all objects in helm chart

BenGalewsky:
Fixes #861 
# Problem
Admins will want to be able to specify memory and CPU limits for every pod deployed into the kubernetes cluster. They will also want to be able to specify the initial request. [This article](https://www.sysdig.com/blog/kubernetes-limits-requests#introduction-to-kubernetes-limits-and-requests) explains the basic concepts and how to determine optimal settings.

Currently there is limited scope to set these for the ServiceX pods.

# Approach
I added properties to values.yaml to set these for each component in the deployment. Each CodeGen can have their own distinct values. There is one set of values applied globally for all of the data lifecycle scheduled tasks. Typically these tasks don't do much on their own so they shouldn't need fine-grained settings.

The TransformerManager already knew how to set limits for the workers, so I just added settings to control the request.

Finally, I took this opportunity to pay down a little technical debt and cleaned up some ugly duplicated code in the TransformerManager unit tests. These changes are in their own commit.

This PR is divided into three commits:

1. Added values and yaml to specify memory and CPU limits/requests for everything in helm deployment
2. The transformer manager was already setting limits for the workers. Added CPU and memory requests to the config and applying those
3. The transformer manager had lots of duplicated config maps - made a fixture that distributes common config

# How to Test
You can change the settings of the different pods in your local values.yaml and then describe the pod to see that they are applied.


MattShirley (approved):

MattShirley:
Looks good. I loaded it on my local cluster fine and was able to observe the resource requests/limits.

PR #1291, opened 2026-01-30
Pod Annotations for everything

### LINKED ISSUE #1132: Add Helm parameters for more configurable k8s pod scheduling
I would like to be able to specify `nodeSelector` and `tolerations` to steer pods to specific nodes in a cluster. Other parameters useful for scheduling could be `annotations` and `affinity`.

To start, this can be implemented as global parameters to apply to all pods. But there may be a case for specifying them differently for different ServiceX components (e.g. use a specific type of "worker" nodes to scale transformer pods).

BenGalewsky:
Fixes #1132 

# Problem
Some kubernetes clusters have nodes with different capabilities. It is useful to be able to direct different pods to specific nodes. For example on River, only nodes with `cvmfs` attribute have the cvmfs node mount.

We might want certain science containers to be directed to nodes with GPUs.

# Approach
Added settings for:
- nodeSelector                                                                                                                                                                  
- tolerations                                                                                                                                                                   
- annotations                                                                                                                                                                   
- affinity   

To all of the microservices and apply them in the templates.

Added similar values for the transformers. These are included in the app configmap and used by the `TransformerManager` to apply to the generated pods

# How to test
Deploy to your cluster using the helm chart from this branch.

Set your app image to be tagged `1132_helm_annotations`


kondratyevd:
Thanks @BenGalewsky, the node selectors and tolerations work well for components where they are enabled. 

In addition to these new parameters, I also needed to add node selectors and tolerations under `minio`, `rabbitmq` and `postgresql.primary` sections of the `values.yaml` to make sure that all ServiceX pods run on correct nodes.

kondratyevd (approved):

PR #1295, opened 2026-02-05
Add Kibana log link to the requests table dashboard

### LINKED ISSUE #1111: Links to log messages on the dashbaord
# Story
_As a user of Servicex I want to see a link to the elastic search results for a transform so I can quickly see what errors occurred when I see not all files completed on a transform_

# Assumptions
1. I have a transform in the dashboard.

# Acceptance Criteria
1. There is a link on the dashboard home page or the transform details page that will take me to a elastic search page that has messages for that transform. This link should be similar to the link provided by the frontend after a crash.

BenGalewsky:
Fixes #1111 

Adds a clickable log icon next to each transformation request that links directly to Kibana with pre-filtered query parameters for the request ID and log level. Also added a link from the transform detail web page.

During testing, I realized the text substitution [used in the front end code](https://github.com/ssl-hep/ServiceX_frontend/blob/e761557dcea06a501682193d3d6bacec84dfd716/servicex/app/transforms.py#L265)  only works with the defaults on the Kibana dashboard and breaks for local or the river deployments.

After several rounds of trying to come up with a way to parse the URL I concluded we needed a friendlier base URL. I updated the [Log url in `values.yaml`](https://github.com/ssl-hep/ServiceX/blob/e593fe3c553de4f68171388a2b3ce6bebd5b46b1/helm/servicex/values.yaml#L293).

This will require a similar change to the frontend to make the error links work correctly.

This will also mean that this is a breaking change for that functionality in the front end.  See [Front end PR 717](https://github.com/ssl-hep/ServiceX_frontend/pull/717)


servicex_app/servicex_app/web/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class LogLevel(str, Enum):
+    r"""
+    Level of the log messages: INFO & ERROR
+    This controls the log level of the logs displayed in the dashboard
+    """
+
+    info = ("INFO",)
+    error = ("ERROR",)
```
MattShirley: Why are these tuples? Shouldn't they just be strings? (Because of Python tuple unpacking, the class attributes are being set as expected, but this is unclear.)

servicex_app/servicex_app/web/__init__.py
```diff
@@ -0,0 +1,39 @@
+from enum import Enum
+from flask import current_app
+import re
+
+
+class LogLevel(str, Enum):
```
MattShirley: Nitpick: is this class necessary? You can use `logging.getLevelName` to convert `logging.INFO` etc. to a human readable value (as needed below).

servicex_app/servicex_app/web/__init__.py
BenGalewsky: Ah, these are in the implementation in the frontend library. I just cloned without thinking... I fixed this here

servicex_app/servicex_app/web/__init__.py
BenGalewsky: These values are from a different domain. They are valid values in the ElasticSearch db and not necessarily related to the Python logger

MattShirley (approved):

helm/servicex/values.yaml
```diff
@@ -289,7 +289,8 @@ logging:
     port: 5959
     protocol: TCP
     monitor: "https://atlas-kibana.mwt2.org:5601/s/servicex/app/dashboards?auth_provider_hint=anonymous1#/view/c2cc1f30-4a5b-11ed-afcf-d91dad577662?embed=true&_g=(filters%3A!()%2CrefreshInterval%3A(pause%3A!t%2Cvalue%3A0)%2Ctime%3A(from%3Anow-24h%2Fh%2Cto%3Anow))&show-time-filter=true"
-    logs: "https://atlas-kibana.mwt2.org:5601/s/servicex/app/dashboards?auth_provider_hint=anonymous1#/view/bb682100-5558-11ed-afcf-d91dad577662?embed=true&_g=(filters%3A!(('%24state'%3A(store%3AglobalState)%2Cmeta%3A(alias%3A!n%2Cdisabled%3A!f%2Cindex%3A'923eaa00-45b9-11ed-afcf-d91dad577662'%2Ckey%3Ainstance%2Cnegate%3A!f%2Cparams%3A(query%3Aservicex)%2Ctype%3Aphrase)%2Cquery%3A(term%3A(instance%3A{{ .Release.Name }}))))%2CrefreshInterval%3A(pause%3A!t%2Cvalue%3A0)%2Ctime%3A(from%3Anow-24h%2Fh%2Cto%3Anow))&show-query-input=true&show-time-filter=true&hide-filter-bar=true"
+    # This URL format is tied to a parser in servicex_app/servicex_app/web/kibana_url_filter.py
```
ponyisi, line 292: Will anyone ever have to change this URL (I guess this is linked to "everyone is using the Chicago Kibana instance" right?)

ponyisi (approved):
Probably the minimal that works without some real refactoring of the Helm chart

helm/servicex/values.yaml
BenGalewsky: Correct - Somebody someday may want to deploy their own ElasticSearch

BenGalewsky:
@ponyisi shall I wait for a final review from you?

ponyisi (approved):

Issue #1304, opened 2026-02-07
Update/Change overall structure of documentation.


RogerJanusiak:
In order to make the documentation fit best in the ServiceX website I think that we should separate the developer guide from the deployment guide. I also think that we should move the documentation to markdown/sphinx in a way that is consistent with all of documentation.

New form of documentation should fit ServiceX documentation standards.

RogerJanusiak:
After talking about this at the Austin retreat we have decided that the documentation should not be put on the main website. I think there is then the argument to just have the documentation as markdown files in the repo so that we don't have to host it and people don't need to google it to find it like right now. These docks are not as generally needed to be accessed constantly like the others!

PR #1312, opened 2026-02-11
Simplified docs by not using mkdocs and moving things to simple .md files.

RogerJanusiak:
Based on our converstations this weekend out how we are going to manage the docs we had discussed making the Backend docs localized to the repo. This is my first pass at doing that, just moving all the docs into markdown files.

There might be ways that do this in a better looking organized way, like the docs used to do, but I don't think that these are being used enough to need to maintain the additional overhead. This should be an easy enough way to get the information across.


RogerJanusiak:
Addressing issue #1304

BenGalewsky:
I'm not sure merging the markdown files is such a good idea. It's pretty hard to find things now. For example, I use the `reference.md` a lot to see the various deployment options.

Maybe keep the files as they are, but make a new table of contents in `docs/README. md` - of course the biggest benefit of makekdocs is maintaining a table of contents without manual intervention..

These docs change pretty slowly, so I think it would be ok to manually manage the index

RogerJanusiak:
Will unmerge the files.

RogerJanusiak:
@BenGalewsky I made changes based on your feedback. If you go into the readme in the docs folder it links to everything through a table of contents.

BenGalewsky (approved):
Yes, thanks

PR #1323, opened 2026-03-04
add basic admin page with reporting

### LINKED ISSUE #1172: REST Endpoint To Report Users and Transform Statistics
_As a ServiceX admin I want to be able to report user metrics so I can monitor uptake of the service_

1. Requires the ability to create admin users
2. Create CSV output

MattShirley:
Adds a basic Admin page.

Screenshot 2026-03-27 at 9 52 14 AM

There are currently two main features on the admin page:

1. Interact with user objects via standard `flask_admin.contrib.sqla.ModelView` based `UserModelView` (including toggling admin/pending status).
2. Run reports.

Additionally, there are a few additional changes made to the CLI commands:
1. `UserModel.sub` has an index added to support the existing `UserModel.find_by_sub` method.
2. The existing `flask user list` command has an admin field added to the output.
3. The existing `flask user list` command has an optional `--email-filter` added.
4. `make-admin` and `revoke-admin` CLI functions have been added using `sub` lookup per Peter's request.

Because the originating ticket didn't have a ton of clarity on what kind of reports we would eventually like to add, I interpreted this to indicate we knew we'd want reporting functionality, but didn't know what specific reports we might need in the future. Thus the core `ReportView` and `SqlCsvReportView` are intended to be flexible and easy to extend. You can see `servicex_app/servicex_app/web/admin/reports/user_transformation_count.py` for an example of a class/report implementing an actual report (count of transforms run by each user over a period of time).

Perhaps most importantly, because `flask-admin` is just a web view that is *only* enabled when `ENABLED_AUTH=True`, the reporting functionality needs to be available from the `flask` CLI as well. The core `ReportView` implements these CLI commands automatically for all subclasses of `ReportView`. So the following is built automatically for every report:

```bash
$ pwd
/home/servicex/servicex_app
$ flask --app 'app:app' reports
Usage: flask reports [OPTIONS] COMMAND [ARGS]...

Options:
  --help  Show this message and exit.

Commands:
  user-transformation-count
$ flask --app 'app:app' reports user-transformation-count --help
Usage: flask reports user-transformation-count [OPTIONS]

Options:
  --days INTEGER  Number of days to look back
  --help          Show this message and exit.
$ flask --app 'app:app' reports user-transformation-count
Name,Email,Institution,Experiment,Transforms (Last 30 Days)
Matthew Scott Shirley,mattutshirley@gmail.com,CERN,ATLAS,1
```


BenGalewsky:
@MattShirley can you explain more about how you handled admin users? How are they created? It looks like there is a new scope in the JWT, can you provide more detail on that?

Also, can you resolve the current conflict with the `develop` branch?

MattShirley:
@BenGalewsky if you're referring to the `session["next"]` flow, you're right, that's not needed and left over cruft from earlier in the PR. I'm removing it.

As far as creating an admin, right now the idea is to use the built in model editor provided by `flask-admin`: 
Screenshot 2026-03-30 at 2 22 07 PM

I wasn't planning on adding an additional CLI command to this PR for creating/setting admins as the admin page will currently always be available instead. However, I can add such a CLI command if you want (and even make the admin page optional).

I'll clean this up and request reviews by the end of the day.

ponyisi:
@MattShirley do you consider this complete?

MattShirley:
> @MattShirley do you consider this complete?

Yes, I just added a short doc page to point users to the admin page.

The only other feature I can think of is to be able to disable the admin page entirely via helm/env config.

ponyisi (approved):
Let's go

Issue #1317, opened 2026-02-18
Race condition in `submit.py:post`


ponyisi:
We have, in https://github.com/ssl-hep/ServiceX/blob/0b7f40c2890d538f702d95af1a55363a49f2dd91/servicex_app/servicex_app/resources/transformation/submit.py#L258-L268 , a check on `dataset_manager.is_lookup_required` and then potentially a lookup request submission which changes the value of that property to prevent further lookups. It seems that it is currently possible for multiple processes to hit this check simultaneously for a single dataset and trigger multiple lookups (which corrupts the file records). This block needs to do a row lock on the relevant dataset.

Issue #1316, opened 2026-02-18
Transaction Leak in dataset_manager


BenGalewsky:
# Problem
There is no transaction protection in the [check for whether a dataset currently exists](https://github.com/ssl-hep/ServiceX/blob/0b7f40c2890d538f702d95af1a55363a49f2dd91/servicex_app/servicex_app/dataset_manager.py#L57) so it is possible to create the dataset multiple times

# Approach
Add explicit locking to this section

PR #1324, opened 2026-03-04
Fix transaction leak in transformation submit endpoint (#1317)

BenGalewsky:
Previously the submit endpoint opened implicit SQLAlchemy transactions the moment it touched any model attribute, then held those transactions open across slow external I/O (HTTP calls to the code-generation service and the Docker registry).  This caused unnecessary lock contention and left connections checked out from the pool for the full duration of every submit request.

Root cause
----------
`SubmitTransformationRequest.post()` performed the following steps inside a single, implicit transaction:

1. Created/fetched the Dataset record and called `dataset.save_to_db()`.
2. Built a `TransformRequest` object, then called `request_rec.save_to_db()`.
3. Made an HTTP POST to the code-gen service (`generate_code_for_selection`).
4. Optionally made another HTTP call to validate the transformer Docker image.
5. Called `db.session.commit()`.

Steps 3–4 could take several seconds; the DB connection (and any row-level locks) was held for that entire window.

Fix
---
Restructured the submit flow so that all external I/O happens *before* any database transaction is opened:

1. **Code generation first (no transaction)** – Call `generate_code_for_selection` before touching the DB.  The method signature was also simplified: instead of receiving a full `TransformRequest` ORM object, it now accepts the two primitive values it actually needs (`selection_string` and `request_id`), removing the hidden coupling to the model and making it possible to call this outside of the DB Transaction.

2. **Docker image validation (no transaction)** – The image-existence check remains outside any transaction.

3. **Object-store bucket creation (no transaction)** – Unchanged in position; still outside the transaction block.

4. **Transaction 1 – persist Dataset** – A short `with session.begin()` block calls `_initialize_dataset_manager`, which now calls `db.session.add(dataset)` instead of `dataset.save_to_db()`.  The context manager commits immediately on exit, assigning the auto-increment `dataset.id` before it is referenced by the TransformRequest.

5. **Transaction 2 – persist TransformRequest** – A second `with session.begin()` block builds and adds the `TransformRequest` using all values already computed in step 1 (image, language, command, configmap name).  Status routing (lookup / running / pending_lookup) and the Celery DID-finder submission also happen inside this transaction, so the request is never visible to other readers in an inconsistent state.

`dataset_manager.py` changes
-----------------------------
Removed the individual `save_to_db()` calls that were scattered through `from_did` and `from_file_list`.  Both paths now call `db.session.add(dataset)` once at the end of the method, letting the *caller's* transaction boundary control the commit.

Test updates
------------
* Updated `test_code_gen_adapter.py` to pass `selection_string` and `request_id` as keyword arguments instead of a `TransformRequest` object.
* Updated `test_dataset_manager.py` to wrap `from_did` / `from_file_list` calls in `with db.session.begin()` blocks, matching the new requirement that callers own the transaction.
* Fixed two timezone-aware comparison assertions that were failing with Python's `datetime.fromtimestamp(0)` (now uses `tz=timezone.utc`).
* Fixed three submit tests (`test_submit_transformation_bad_did_scheme`, `test_submit_transformation_missing_dataset_source`, `test_submit_transformation_duplicate_dataset_source`) that failed because code-gen is now called before the dataset path, so a `mock_codegen` fixture is required even for tests that exercise dataset-validation errors.

### DIFF
--- a/servicex_app/servicex_app/code_gen_adapter.py
+++ b/servicex_app/servicex_app/code_gen_adapter.py
@@ -27,8 +27,6 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import requests
 from requests_toolbelt.multipart import decoder
-
-from servicex_app.models import TransformRequest
 from servicex_app.reliable_requests import REQUEST_TIMEOUT, servicex_retry
 
 
@@ -43,13 +41,18 @@ class CodeGenAdapter:
         return result
 
     def generate_code_for_selection(
-        self, request_record: TransformRequest, namespace: str, user_codegen_name: str
+        self,
+        selection_string: str,
+        request_id: str,
+        namespace: str,
+        user_codegen_name: str,
     ) -> tuple[str, str, str, str]:
         """
         Generates the C++ code for a request's selection string.
         Places the results in a ConfigMap resource in the
         Starts a transformation request, deploys transformers, and updates record.
-        :param request_record: A TransformationRequest.
+        :param selection_string: String representing the selection criteria.
+        :param request_id: ID of the request.
         :param namespace: Namespace in which to place resulting ConfigMap.
         :param user_codegen_name: Name provided by user for selecting the codegen URL
                from config dictionary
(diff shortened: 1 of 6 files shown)


ponyisi (approved):
Looks OK to me.

ponyisi:
Does this also address #1316 ?

Issue #1318, opened 2026-02-18
Possible issue with file size handling on retries


ponyisi:
In the situation that a transform has run successfully but is then retried (due to messaging glitches, etc.) then only the file size from the first run will be recorded in Postgres. However the retries will unconditionally upload results to the object store, and these outputs are not guaranteed to be the same size (e.g. anything that does a computation, like TopCP, may have slight floating point differences between CPUs which cascade in the compression). In this case the final files on S3 may have different sizes from the values in the database, which upsets the client (which expects the two to be the same, at least for `servicex deliver`). In any case we should decide what to do here: should a sidecar check if it's a retry of a successful run and abort, or should we update the DB when there's a second complete file report?

PR #1331, opened 2026-03-08
Accept multiple result reports

ponyisi:
Retrying a transformation may result in an output with a different size (which then breaks file accounting and creates a mismatch with the file in S3). By the time the file is reported to the `servicex_app`, the result has already been uploaded to S3 so this is a fait accompli, so we might as well update the DB record for the file instead of whining about it. Addresses #1318.


servicex_app/servicex_app/resources/internal/transformer_file_complete.py
```diff
@@ -19,7 +19,7 @@
 # AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 # IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
 # DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
-# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# FOR ANY DIresultT, INDIresultT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
```
BenGalewsky, line 22: What the what? Is this a subtle ransom message

BenGalewsky (approved):

PR #1335, opened 2026-03-10
Switch to combined ROOT+uproot image where relevant

ponyisi:
We have a new `pixi`-based science image which allows us to combine ROOT+uproot (and also includes a few other things like `pandas`). This enables e.g. `RDataFrame` bridging to awkward arrays. The new science container also reduces maintenance complexity since it is Alma 9-based unlike the Ubuntu-based uproot-only images.


BenGalewsky (approved):
This looks like a nice improvement

PR #1341, opened 2026-03-10
Increase default memory limit on App to avoid OOM errors

BenGalewsky:


ponyisi (commented):
Works. I wonder if this means we need to bump up request to 2 Gi and limit to, say, 3?

ponyisi (approved):
Thanks!

PR #1342, opened 2026-03-11
Fix error in job submission

ponyisi:
#1324 refactored code so that the created `TransformRequest` for a submission was not saved to the database before being used to dispatch available files. However that saving set the values of `files_completed` and `files_failed`, which otherwise are `None`, causing the computation of `files_remaining` in the JSON serialization to fail.


BenGalewsky (approved):
Ah, thank you

PR #1357, opened 2026-03-19
update `poetry.lock`

MattShirley:
Some quality of life changes.

1. Update `poetry.lock`
2. Remove `pkg_resources` calls from code (which `setuptools==^82` removed). Versioning `setuptools` to 81 will also fix this error, but it is simple enough to remove calls to `pkg_resources`.
3. Move `pre-commit` to `dev` group in `pyproject.toml`. Obviously this is good practice, [but not doing so](https://github.com/ssl-hep/ServiceX/actions/runs/23321659118/job/67834537851?pr=1357) will now generate an error in the build step due to a compatibility issue with `virtualenv`. Moving it to the `dev` group fixes this.


BenGalewsky (approved):
Ok - looks like a useful change while we switch to `uv`

MattShirley:
@BenGalewsky yes I will log a ticket for the `uv` change

PR #1371, opened 2026-03-26
Protect against possible race condition if dataset is empty

ponyisi:
It seems that it may be possible for k8s to have enough of a delay creating a transformer deployment that we could try to shut down a deployment for having zero files before the deployment is fully created. In that case we have deleted the configmap but not the deployment, and we have the appearance of a lingering deployment with no configmap, stuck forever in `ContainerCreating`. Add retries to all the shutdown commands; also try to delete the deployment before the configmap.


BenGalewsky (approved):
Knocking these over one by one. Thank you @ponyisi

Issue #1310, opened 2026-02-08
Upgrade to UpRoot 5.7.0


BenGalewsky:
Update codegens to not assume default output is TTree
Will impact func_adl uproot, uproot raw, and python codegens
Will need to see if this impacts get_structure

PR #1378, opened 2026-03-30
Support for uproot 5.7.0, rntuple support and translation

ponyisi:
* Makes tweaks for uproot 5.7.* (so that requested TTree output doesn't produce RNTuple instead)
* Adds root-rntuple output support for FuncADL Uproot, Python transformers
* Adds TTree -> RNtuple translation support

Note that this cleans up places where the string `root` was used incorrectly instead of `root-file`. This would have led to wrong behavior in certain TTree -> parquet conversion edge cases.


ponyisi:
This should resolve #1310 , #999

BenGalewsky:
You should mention in the PR description that we are migrating from `root` to `root-file`
Does this impact the front end? How do we communicate this new output format?

ponyisi:
Hi @BenGalewsky I don't think we're "migrating" from `root` to `root-file`, the actual string as used in the server has been `root-file` for a long time - we're actually cleaning up inconsistencies in the code. For example the Python codegen tests `root-file` and not `root`:
https://github.com/ssl-hep/ServiceX/blob/c29f80b62215db1e55c855dfe0f5691b4d2f4967/code_generator_python/python_code_generator/templates/transform_single_file.py#L48
and the option when you submit is `root-file`:
https://github.com/ssl-hep/ServiceX/blob/c29f80b62215db1e55c855dfe0f5691b4d2f4967/servicex_app/servicex_app/resources/transformation/submit.py#L111

It's just that the `transformer_capabilities.json` files weren't set up properly. (Main lesson from this is that the format transformation capability of the sidecars was clearly not really used or it would have been spotted earlier.)

kyungeonchoi (approved):
Thank you! Looks good to me.

A minor opinion: would it be good place to change `root-file` to `root-tree`?

ponyisi:
@kyungeonchoi I would rather not do a renaming of `root-file` to `root-ttree` since this is part of the interface to the client (sadly) - it's the string sent in the request. It's not impossible but it would be delicate.

PR #1382, opened 2026-04-05
Update did_finder_rucio for security issue

ponyisi:
Updates requests to a version >= 2.33. In the process forces a move to Python >= 3.10 which was overdue anyway.


ponyisi:
This should properly be fixed by an update to the upstream container we build from. Closing

PR #1383, opened 2026-04-05
Update x509 finder for security issue

ponyisi:
Requests >= 2.33.0 and Python 3.10.


ponyisi:
This should properly be fixed by an update to the upstream container we build from. Closing

PR #1385, opened 2026-04-08
Stop `pending_lookup` requests if dataset bad

ponyisi:
A bad dataset error should scrub pending transformations as well as running ones. (I originally wrote the code thinking that pending_lookup transforms did not get pods started but I was wrong there.)


kratsg:
Nice job :) Let's hope this fixes what we see on the AF.

BenGalewsky (approved):
Oof, good catch

PR #1403, opened 2026-04-16
Implement the row lock to prevent multiple lookups for the same dataset

ponyisi:
#1317 refactored transactions for the dataset lookup, but it did not implement an actual lock, so it did not fix the race condition. Add a method to force the issuance of a `SELECT ... FOR UPDATE` in the relevant place.


ponyisi:
Hopefully actually fixes #1316

MattShirley (approved):

PR #1412, opened 2026-04-23
log client version ID on server side

### LINKED ISSUE #832: Track version numbers of client
Accept client version numbers at some point in a connection and keep track of the values. This will allow us to know if there are still significant numbers of users using deprecated versions.

Implement this as a DEBUG level log message upon connect that shows the client version

MattShirley:
Logs client side client version when received as an HTTP header.

For issue: https://github.com/ssl-hep/ServiceX/issues/832

Front end work: https://github.com/ssl-hep/ServiceX_frontend/pull/736

### DIFF
--- a/servicex_app/servicex_app/__init__.py
+++ b/servicex_app/servicex_app/__init__.py
@@ -35,7 +35,7 @@ from distutils.util import strtobool
 import base64
 import click
 import logstash
-from flask import Flask, current_app, request
+from flask import Flask
 from flask.cli import AppGroup
 from flask_bootstrap import Bootstrap5
 from flask_cors import CORS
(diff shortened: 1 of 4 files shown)


BenGalewsky (approved):
Ok - this is a good start. I guess next step is to record this somewhere

servicex_app/servicex_app/__init__.py
```diff
@@ -387,4 +387,10 @@ def inject_modules():
                 create_kibana_link=create_kibana_link,
             )
 
+    @app.before_request
```
BenGalewsky: This seems like it is unnecessarily frequent. Couldn't we just do this on the Login message?

MattShirley: What would be the login route for the client? We could log when a token refresh happens or a transformation is submitted.

BenGalewsky: Exactly - so that's pretty much once per transform request and not when polling for results

MattShirley: I've moved the global logger to only occur when a transformation is submitted, will update the client as well.

PR #1415, opened 2026-04-23
Fixes bugs introduced in #1378

ponyisi:
* report number of events in Python transformer regardless of TTree/RNTuple code path
* get correct file size if file is converted


BenGalewsky (approved):
Ok, thanks

PR #1431, opened 2026-05-11
Fix deadlock of #1403

ponyisi:
PR #1403 introduced a potential deadlock due to the presence of multiple transactions affecting a single dataset in the `submit.py` code path. Put those both in a single transaction. Also add Postgres upsert logic to ensure non-stale entries in `datasets` are unique, so we only create one dataset entry per DID even if multiple requests are made in a short time.


ponyisi:
Test failures are due to a mismatch of postgres/sqlite dialects, looking into this

BenGalewsky (approved):
Nice one!

PR #1438, opened 2026-05-12
Server-side download URL generation

ponyisi:
This is a first pass at generating download URLs for ServiceX output on the server side. The client can request a list of files in a transform, and URLs will be returned.

We have a hook here for doing this with other backends, e.g. xrootd-based ones. At some point we also probably want to cache results for some length of time.


servicex_app/servicex_app/routes.py
```diff
@@ -156,6 +157,7 @@ def add_routes(
     prefix = "/servicex/transformation"
     api.add_resource(SubmitTransformationRequest, prefix)
     api.add_resource(AllTransformationRequests, prefix)
+    api.add_resource(FileURLGenerator, prefix + "/file_urls")
```
MattShirley: The convention is to use dashes in the route paths.

servicex_app/servicex_app/resources/transformation/file_urls.py
```diff
@@ -0,0 +1,96 @@
...
+from servicex_app.decorators import auth_required
+from servicex_app.models import TransformRequest
+from servicex_app.resources.servicex_resource import ServiceXResource
+
+import boto3
+
+
+class FileURLGenerator(ServiceXResource):
+    def __init__(self):
+        if True:  # S3 backend
```
MattShirley: Why is this conditional block needed? How does S3 hook in to this?

servicex_app/servicex_app/routes.py
ponyisi: Fixed this

servicex_app/servicex_app/resources/transformation/file_urls.py
ponyisi: Sorry, this is admittedly unclear. Basically, this is a placeholder for an if-statement for a different storage backend (e.g. xrootd).

MattShirley: Can you set a `backend` class attribute that is typed as a literal and defaults to "s3"? Then it can be used for the two conditional blocks.

MattShirley: I guess it actually shouldn't be a class attribute but an instance attribute so it's defined when the object is created

MattShirley: Actually thinking about this further, the backend is a global setting right (i.e. we don't support multiple backends)? Then the backend system should be defined in the helm values and can be used in the conditional here.

But also, when does the `__init__` method fire for flask routes like this? I don't see it commonly being used for resources (and there's no super call). Is one object created when the route is setup, or is one object created for every request? If the latter, then this `__init__` implementation would generate unnecessary boto objects. Instead, the client could be moved outside FileURLGenerator so one client is used consistently.

ponyisi: Yes, the intention was that when we actually figure out how we want to signal the backend system being used via the helm chart, that the conditional would work based on that. 

The `__init__` is, as far as I know, only called when the routes are being set up. I think the call to the superclass constructor is essentially a noop but I'll add it.

servicex_app/servicex_app/resources/transformation/file_urls.py
```diff
@@ -0,0 +1,97 @@
...
+from servicex_app.models import TransformRequest
+from servicex_app.resources.servicex_resource import ServiceXResource
+
+import boto3
+
+
+class FileURLGenerator(ServiceXResource):
+    def __init__(self):
+        super().__init__()
+        if True:  # S3 backend
```
BenGalewsky: I think this unused conditional is a bit ick - It's easy enough to indent this once we implement the next back end. For now this is just confusing

BenGalewsky (approved):
Cheers

PR #1448, opened 2026-05-19
remove need to specify app for flask CLI and update documentation

MattShirley:
Fix the FLASK_APP env variable definition (managed with helm). This will allow the flask CLI including `flask reports` to work without needing to specify an app location.


ponyisi (approved):
Very helpful, thanks!

PR #1455, opened 2026-05-21
add cancel all button to dashboard

### LINKED ISSUE #1296: Add Cancel All Button to Dashboard
While doing some work for the integration test @ArturU043 found that he was going through and canceling jobs on the dashboard one by one. 

Action:
Add a button that allows the user to cancel all of their current transforms.

MattShirley:
Adds Cancel All button to the Dashboard when ENABLE_AUTH=True. This will cancel all active transformations based upon the existing individual Cancel logic.

When ENABLE_AUTH=False, the button is not present and the route returns 400.


servicex_app/servicex_app/resources/transformation/cancel_all.py
```diff
@@ -0,0 +1,81 @@
...
+class CancelAllTransform(ServiceXResource):
+    @classmethod
+    def make_api(cls, transformer_manager: TransformerManager):
+        cls.transformer_manager = transformer_manager
+
+    @auth_required
+    def get(self):
+        if current_app.config.get("ENABLE_AUTH"):
+            user = self.get_requesting_user()
+            user_id = user.id if user is not None else None
```
ponyisi: How does the None case propagate?

servicex_app/servicex_app/resources/transformation/cancel_all.py
```diff
@@ -0,0 +1,81 @@
...
+            transform_reqs = TransformRequest.query.filter(
```
ponyisi: I feel Ben always likes these to be added to the model class so they can be reused easily

servicex_app/servicex_app/resources/transformation/cancel_all.py
```diff
@@ -0,0 +1,81 @@
...
+                TransformRequest.status.in_(_ACTIVE_STATUSES),
+            ).all()
+        else:
+            return {"message": "This is not available when auth is disabled"}, 400
+        canceled_ids = []
+        now = datetime.now(tz=timezone.utc)
+        namespace = current_app.config["TRANSFORMER_NAMESPACE"]
+        for transform_req in transform_reqs:
+            request_id = transform_req.request_id
+            if transform_req.status in (
```
ponyisi: Possible race condition if something moves from pending_lookup to running before cancellation?

servicex_app/servicex_app/resources/transformation/cancel_all.py
```diff
@@ -0,0 +1,81 @@
...
+_ACTIVE_STATUSES = [s for s in TransformStatus if not s.is_complete]
+
+
+class CancelAllTransform(ServiceXResource):
+    @classmethod
+    def make_api(cls, transformer_manager: TransformerManager):
+        cls.transformer_manager = transformer_manager
+
+    @auth_required
+    def get(self):
```
BenGalewsky: Not to be "that guy", but this should be a POST and not a GET since it is taking an action...

MattShirley: Yeah you're right, I copied the pattern from the existing `cancel` route.

MattShirley: Both Cancel and Cancel All were updated to use post.

servicex_app/servicex_app/resources/transformation/cancel_all.py
MattShirley: Two class methods were added to support this.

servicex_app/servicex_app/resources/transformation/cancel_all.py
MattShirley: It now explicitly throws a 400 error.

servicex_app/servicex_app/resources/transformation/cancel_all.py
ponyisi: Hm, can `cancel` be called by the frontend CLI?

BenGalewsky: Yes, it can... I'm not sure it's worth making a breaking change over. Maybe keep cancel endpoint a GET and make CancelAll a POST.

Good catch, @ponyisi

MattShirley: I've reverted the cancel to use get.

Could we support the same logic on both get and post and then deprecate the get on some timeline?

ponyisi: Sounds reasonable. (Sounds like a feature flag. :-) )

MattShirley: @ponyisi @BenGalewsky this has been finalized, with an additional frontend PR added to support it.

servicex_app/servicex_app_test/test_models.py
```diff
@@ -160,3 +166,63 @@ def test_threshold_reached_handles_none(self, app_context):
             # Assert
             assert result is None
             mock_conn.execute.assert_called_once()
+
+
+class TestShutdownPod:
```
BenGalewsky: Why are these tests here, and not in test_transformer_manager.py?

servicex_app/servicex_app/models.py
```diff
@@ -246,6 +251,35 @@ def to_json(self):
...
+        return (
+            cls.query.filter(
+                cls.submitted_by == user.id,
+                cls.status.in_(active_statuses),
+            )
+            .with_for_update()
+            .all()
+        )
+
+    def shutdown_pod(self, transformer_manager: TransformerManager):
```
BenGalewsky: This doesn't belong in the database models. I think it should be an operation in transformer_manager.py

servicex_app/servicex_app/resources/info.py
```diff
@@ -37,5 +37,6 @@ def get(self):
             "capabilities": [
                 "poll_local_transformation_results",
                 "long_sample_titles_10240",
+                "post_cancel_transform",
```
BenGalewsky: I find this name confusing... it sounds like it describes something that happens after a cancel (post-cancel-bliss). Maybe `cancel_transform_post_method` or somthing

servicex_app/servicex_app/models.py
ponyisi: Agree, in my earlier comment I just meant that the DB query (`active_user_transformations`) should be  in the model class

ponyisi (approved):
I think all comments have been addressed here.

PR #1456, opened 2026-05-21
Allow sites to mount CVMFS in transformer pods

ponyisi:
Adds a `values.yaml` hook to enable sites to specify a Kubernetes `volume` with CVMFS to mount inside transformer containers.


servicex_app/servicex_app/transformer_manager.py
```diff
@@ -44,6 +45,25 @@
 
 from servicex_app.models import TransformRequest, TransformStatus
 
+UPPER_FOLLOWED_BY_LOWER_RE = re.compile("(.)([A-Z][a-z]+)")
```
BenGalewsky, line 50: Maybe a comment here about why this is needed?

BenGalewsky:
This looks great. I do wonder if we need more extensive documentation than just a few cells in reference.md

@ponyisi - do you think this is ready to release into the wild? Have you tested it on river?

ponyisi:
@BenGalewsky I think it's good to go but an rc would be really appreciated to fully check the helm chart behavior.

I wonder if we need some kind of broader deployment guide to give something more than just the `values.yaml` documentation?

BenGalewsky (approved):
Ok - let's try this out in the wild! I'll build a pre-release

PR #1457, opened 2026-05-24
Replace slashes by underscores and not colons

ponyisi:
Having colons in the object names causes issues with URL generation and fsspec, since colons have a special meaning in URL parsing and fsspec language. Underscores are much more neutral.


BenGalewsky (approved):
Sounds good

PR #1469, opened 2026-06-03
change sign in redirect page from user dashboard to user profile

### LINKED ISSUE #1298: Login link that after login directs the user to the profile page
It would be nice that the user is able to directed to the profile page where they can copy their token. This would make it easier for new users as well as current users.

MattShirley:
Small PR to change the login redirect to user profile.


BenGalewsky (approved):
Great!

PR #1479, opened 2026-06-15
remove need to log out when `flask user make-admin` is run.

MattShirley:
I observed a "bug" while testing the `flask user make-admin` command. After the CLI command is run, the user does not immediately see the "Admin" panel on the navbar. The user must log out and log back in. (The problem is that the server cannot invalidate a session without installing additional server side session middleware which seems outside the scope of this issue.)

To resolve, this PR checks against the actual db user.admin value every time the `is_admin_user` is run. That function is executed every UI page load and is nested inside `@auth_required` which is most routes. This seems kind of heavy overhead to add given how rarely the `make-admin` CLI command will be run. I think an acceptable (and simpler) alternative would be to add output to the `make-admin` command to instruct the user to log out and log back in. Please share thoughts on whether these changes should be made.

### DIFF
--- a/servicex_app/servicex_app/decorators.py
+++ b/servicex_app/servicex_app/decorators.py
@@ -11,7 +11,7 @@ from flask import (
     url_for,
 )
 from flask_jwt_extended import get_jwt_identity, jwt_required, verify_jwt_in_request
-from flask_jwt_extended.exceptions import NoAuthorizationError, UserClaimsVerificationError
+from flask_jwt_extended.exceptions import NoAuthorizationError
 
 from servicex_app.models import UserModel, db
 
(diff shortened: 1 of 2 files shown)


BenGalewsky:
no, we certainly don't need to add a db hit for every page just for this very rare case. Please add the instruction to the command.

I assume this also impacts when an admin creates a new admin thorough the web front end?

MattShirley:
> no, we certainly don't need to add a db hit for every page just for this very rare case. Please add the instruction to the command.
> 
> I assume this also impacts when an admin creates a new admin thorough the web front end?

Yes, I would assume so since by default (without a server side session library), the session lives in the browser.

PR #1480, opened 2026-06-15
admin page and cli improvements

MattShirley:
Makes the following changes:

1. Change user CLI commands to use email instead of sub.
2. Make `AdminAuthMixin` lookup the `user.admin` value in the db instead of session.
3. Add a line to the `make-admin` CLI output instructing the user to log out and back in.


BenGalewsky (approved):
ok -looks like an improvement

PR #1491, opened 2026-06-24
Fix requests table horizontal scrolling and truncate long titles

### LINKED ISSUE #1299: Dashboard table scaling on Safari
I have found that the dashboard page table doesn't scale nicely on Safari. When I am working in a half window I cannot see the whole table.

BenGalewsky:
Resolves #1299 

# Problem 
This isn't really a safari issue, but an issue with how the dashboard scales when the window is resized.

# Approach
Switch the table wrapper to table-responsive so a horizontal scrollbar appears whenever the table overflows (the old table-responsive-xxl class does not exist in Bootstrap 4.5 and was a no-op). Constrain the Title column with text-overflow: ellipsis instead of word-break: break-all, and add a hover tooltip showing the full title.

Bound the dashboard content wrappers with col-12 so the table-responsive container is capped at the page width; previously the unbounded flex child grew to the table's full width, defeating overflow-x and clipping the right-most columns off-screen.

Resizes nicely now and puts a horizontal scroller if we get too skinny

image

### DIFF
--- a/servicex_app/servicex_app/templates/global_dashboard.html
+++ b/servicex_app/servicex_app/templates/global_dashboard.html
@@ -5,7 +5,7 @@
 
 {% block content %}
 
-  <div>
+  <div class="col-12">
     <div class="content-section">
       <div class="row justify-content-between align-items-center">
         <h4 class="mb-3 col-auto">Transformation Requests</h4>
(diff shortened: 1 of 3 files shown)


RogerJanusiak (approved):
Looks great! Thank you for the fix!

PR #1495, opened 2026-07-02
add basic usage report

### LINKED ISSUE #1478: Download Usage Report as CSV file
_As a ServiceX admin I want to download a CSV file of usage so I can generate my own graphs and reports_

Generated at UIUC workshop

Should only be accessible to admin users

Add button to the admin page to download a CSV usage report. 

Join submitted_by to the user object and include the email, institution,  name and created_at columns from that table.

Note to developer: Make a quick calculation of the approximate size per row and use that to determine a LIMIT on this report to keep it under 200Mb. Sort by date, descending so we always see the most recent transform requests even after the table gets really large

MattShirley:
Instead of limiting this report based on SQL rows as the Issue requested, I have added logic to `SqlCsvReportView` to allow a max size to be added to a report. Output will be truncated by row if the max download size is exceeded.

### DIFF
--- a/servicex_app/servicex_app/web/admin/reports/__init__.py
+++ b/servicex_app/servicex_app/web/admin/reports/__init__.py
@@ -26,7 +26,7 @@ class ReportView(AdminAuthMixin, BaseView):
         if "abstract" not in cls.__dict__:
             cls.abstract = False
 
-    def write_output(self, output: TextIO, **kwargs):
+    def write_output(self, output: io.StringIO, **kwargs):
         raise NotImplementedError
 
     @expose("/")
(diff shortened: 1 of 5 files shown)


servicex_app/servicex_app/web/admin/reports/__init__.py
```diff
@@ -26,7 +25,7 @@ def __init_subclass__(cls, **kwargs):
         if "abstract" not in cls.__dict__:
             cls.abstract = False
 
-    def write_output(self, output: TextIO, **kwargs):
+    def write_output(self, output: io.StringIO, **kwargs):
```
ponyisi, line 28: Curious why this is changed

ponyisi (commented):
See the one thing, otherwise looks good to me

servicex_app/servicex_app/web/admin/reports/__init__.py
MattShirley: The downstream `SqlCsvReportView` needs it to be `StringIO` to access the `getvalue` method while implementing the max download size. I wanted to keep consistency with the class hierarchy.

servicex_app/servicex_app/web/admin/reports/usage_report.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def get_query(self) -> Select:
+        return (
+            select(
+                UserModel.name.label("Name"),
+                UserModel.email.label("Email"),
+                UserModel.institution.label("Institution"),
+                TransformRequest.submit_time.label("Run Time"),
+            )
+            .join(TransformRequest, TransformRequest.submitted_by == UserModel.id)
```
MattShirley, line 39: Hmm.. I wonder if there still needs to be a limit here in cases of large TransformRequest tables. How much memory do pods usually have in prod?

ponyisi: We would expect 2-3 GB?

MattShirley: I changed `SqlCsvReportView` to always stream results from the SQL server in 1000 row increments, so this shouldn't be a problem for general reporting. Expected large CSVs can just implement the `max_download_size` (which we can optionally add a default value, currently it's unbounded by default).

ponyisi:
As discussed in the meeting earlier today - it might be nice to just dump everything present in the `TransformRequest` model (aka the `requests` table) instead of trying to be clever and cherry-pick?

MattShirley:
> As discussed in the meeting earlier today - it might be nice to just dump everything present in the `TransformRequest` model (aka the `requests` table) instead of trying to be clever and cherry-pick?

@ponyisi I updated the field list to include most fields on TransformRequest. I have still excluded some fields, such as database keys, the selection itself, and some of the longer fields that I wasn't sure would be useful. Can you take a look at that and tell me if you want it modified?

BenGalewsky (approved):
yay, thank you

PR #1496, opened 2026-07-02
[codex] funcadl uproot rntuple parity

gordonwatts:
## What changed
- Refactored the func_adl uproot ROOT writer to use a shared helper for `root-file` and `root-rntuple` output.
- Added tests for the writer contract and an empty-array `root-rntuple` transform path.

## Why
- This brings the func_adl generator in line with the raw uproot generator’s output behavior, especially for RNTuple handling.

## Validation
- Local execution of Python tests was not possible in this environment because only the Windows Store `python.exe` shim is available here.
- Changes were reviewed statically and committed as a focused two-file patch.


BenGalewsky:
@gordonwatts do you want this in the 1.8 release? The clock is ticking...

PR #1501, opened 2026-07-08
build: migrate servicex_app from Flask 2 to Flask 3

BenGalewsky:
Bump flask ^2.3.3 -> ^3.1 (resolves to 3.1.3). Werkzeug, Jinja2, and all flask-* extensions were already at Flask-3-compatible versions; flask-restful 0.3.10 verified working against Flask 3.1.

Related cleanups:
- Bump flask-migrate ^3.1.0 -> ^4.1 and pytest-flask ^1.2.0 -> ^1.3.0
- Replace distutils.util.strtobool (removed in Python 3.12) with a local equivalent
- Fix Api(app, errors=Flask.errorhandler): the errors kwarg expects a dict of exception-name -> response specs; passing the unbound method was a latent bug that would raise TypeError inside flask-restful error handling

Assisted-by: ClaudeCode:claude-fable-5


ponyisi:
One comment before I look more closely - can we get a test for full coverage of `strtobool` ?

ponyisi:
Less invasive than I expected. Have you checked that `alembic` doesn't think a migration is needed (there seem to have been some strictness updates)?

BenGalewsky:
@ponyisi - Tests added to bring coverage back up.
I've played around some with the migrations and it seems like we are working ok...

BenGalewsky:
> Less invasive than I expected.

Yes, I was girding myself for a major migration

ponyisi (approved):

PR #1516, opened 2026-08-03
feat(webui): collapse long Selection display with show more/less toggle

BenGalewsky:
The Selection field on the Transformation Request page can be very large, pushing the rest of the definition list far down the page. Clamp it to a few lines by default and add a browser-side toggle to expand/collapse the full value.

Assisted-by: ClaudeCode:claude-opus-4.8


RogerJanusiak:
Took a second to be able to test it, looks great!

RogerJanusiak (approved):

PR #1520, opened 2026-08-03
hide INFO, DEBUG logs from kibana by default

### LINKED ISSUE #876: Reduce verbosity of routine messages
We should have a campaign to reduce the verbosity of routine messages so that they don't clutter user output. We do not need to see 3-4 messages for every file uploaded to the object store, or most of the task dispatch messages from celery.

Some of this should be implemented by reducing the logging level and updating what shows up in Kabana

MattShirley:
1. Update default log level to send to ES to DEBUG.
2. Update Kibana log url to hide INFO, DEBUG by default

Screenshot 2026-08-03 at 4 25 02 PM

### DIFF
--- a/helm/servicex/values.yaml
+++ b/helm/servicex/values.yaml
@@ -32,7 +32,7 @@ app:
       clusterIssuer: null
       enabled: false
       secretName: "{{.Release.Name}}-app-tls"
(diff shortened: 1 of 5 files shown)


BenGalewsky (approved):
Much better

PR #1530, opened 2026-08-07
Make the home page the dashboard for authenticated users

### LINKED ISSUE #1529: Make "Home" Page the Dashboard
@ponyisi wrote in slack:
>May I make an interface complaint regarding 1.8.5... we got rid of the "Dashboard" link in the top panel, but we don't go directly into the dashboard now when we go to the top-level URL, so it's a bit mysterious how to get it.
And we still have "Welcome to ServiceX" with no other content if you go to https://servicex-release-prod.servicex.ssl-hep.org/
could we perhaps restore the Dashboard link, redirect there automatically if you're logged in, and if you're not, redirect to the login page?

BenGalewsky:
Fixes #1529 
This pull request updates the landing behavior for the home page to provide a more consistent user experience based on authentication status and improves related test coverage. Now, users are redirected to the appropriate dashboard depending on whether authentication is enabled and their sign-in state. The tests have been expanded to cover these scenarios.

**Home page routing improvements:**

* Updated the `home` view in `home.py` to redirect unauthenticated users to the welcome page, authenticated users to their dashboard, and all users to the global dashboard when authentication is disabled.
* Simplified the navigation bar logic in `base.html` by removing conditional links based on authentication status, since routing now handles redirection.

**Test enhancements:**

* Refactored and expanded tests in `test_home.py` to cover all combinations of authentication enabled/disabled and user sign-in state, ensuring correct redirects and template rendering.

### DIFF
--- a/servicex_app/servicex_app/templates/base.html
+++ b/servicex_app/servicex_app/templates/base.html
@@ -42,6 +42,9 @@
           <!-- Navbar Left Side -->
           <ul class="navbar-nav mr-auto">
             <li class="nav-item">
+              <a class="nav-link" href="{{ url_for('home') }}">Dashboard</a>
+            </li>
+            <li class="nav-item">
               <a class="nav-link" href="{{ config['DOCS_BASE_URL'] }}" target="_blank">Docs</a>
             </li>
           </ul>
--- a/servicex_app/servicex_app/templates/base.html
+++ b/servicex_app/servicex_app/templates/base.html
@@ -31,15 +31,7 @@
   <header class="site-header">
     <nav class="navbar navbar-expand-md navbar-dark bg-dark fixed-top">
       <div class="container">
-          {% if config["ENABLE_AUTH"] %}
-            {% if session['is_authenticated'] %}
-            <a class="navbar-brand mr-4" href="{{ url_for('user-dashboard') }}" >ServiceX</a>
-            {% else %}
-            <a class="navbar-brand mr-4" href="{{ url_for('home') }}" >ServiceX</a>
-            {% endif %}
-          {%  else %}
-          <a class="navbar-brand mr-4" href="{{ url_for('global-dashboard') }}" >ServiceX</a>
-          {% endif %}
+          <a class="navbar-brand mr-4" href="{{ url_for('home') }}" >ServiceX</a>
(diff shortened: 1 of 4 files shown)


servicex_app/servicex_app/web/home.py
```diff
@@ -1,5 +1,10 @@
-from flask import render_template
+from flask import current_app, redirect, render_template, session, url_for
 
 
 def home():
+    """Land signed-in users on their dashboard; welcome page for everyone else."""
+    if not current_app.config.get("ENABLE_AUTH"):
+        return redirect(url_for("global-dashboard"))
+    if session.get("is_authenticated"):
```
ponyisi, line 8: suggestion here would be if `ENABLE_AUTH` is set, but user not authenticated, redirect to the sign in

servicex_app/servicex_app/templates/base.html
```diff
@@ -31,15 +31,7 @@
...
-          {% if config["ENABLE_AUTH"] %}
-            {% if session['is_authenticated'] %}
-            <a class="navbar-brand mr-4" href="{{ url_for('user-dashboard') }}" >ServiceX</a>
-            {% else %}
-            <a class="navbar-brand mr-4" href="{{ url_for('home') }}" >ServiceX</a>
-            {% endif %}
-          {%  else %}
-          <a class="navbar-brand mr-4" href="{{ url_for('global-dashboard') }}" >ServiceX</a>
-          {% endif %}
+          <a class="navbar-brand mr-4" href="{{ url_for('home') }}" >ServiceX</a>
```
ponyisi, line 34: Probably a good cleanup. Would still suggest putting back the Dashboard button rdirecting to the user dashboard

ponyisi (commented):
See comments

servicex_app/servicex_app/web/home.py
BenGalewsky: Good bye "Welcome to ServiceX" page

servicex_app/servicex_app/templates/base.html
BenGalewsky: Done

ponyisi (approved):

PR #1534, opened 2026-08-25
save output path on transform model

### LINKED ISSUE #1505: Save transform output path in db
We currently assume we can reconstruct the output path with the transform ID.

Don't require that and save the exact output path in the db

MattShirley:
Adds `output_path` to `requests` table as well as `TransformerManager.compute_output_path` for distinguishing between object store and volume destinations.

### DIFF
--- a/servicex_app/servicex_app/models.py
+++ b/servicex_app/servicex_app/models.py
@@ -194,6 +194,7 @@ class TransformRequest(db.Model):
     workers = db.Column(db.Integer, nullable=True)
     result_destination = db.Column(db.String(32), nullable=False)
     result_format = db.Column(db.String(32), nullable=False)
+    output_path = db.Column(db.String(1024), nullable=True)
     submitted_by = db.Column(db.Integer, db.ForeignKey("users.id"), nullable=True)
 
     files = db.Column(db.Integer, default=0, nullable=False)
@@ -230,6 +231,7 @@ class TransformRequest(db.Model):
             "workers": self.workers,
             "result-destination": self.result_destination,
             "result-format": self.result_format,
+            "output-path": self.output_path,
             "generated-code-cm": self.generated_code_cm,
             "status": self.status.string_name,
             "failure-info": self.failure_description,
--- a/servicex_app/servicex_app/transformer_manager.py
+++ b/servicex_app/servicex_app/transformer_manager.py
@@ -79,6 +79,22 @@ class TransformerManager:
         cls.celery_app = celery_app
         return cls
 
+    @staticmethod
+    def compute_output_path(request_id: str, result_destination: str) -> str:
+        """Return the destination the transformer will write results to.
+
+        For ``object-store`` this is the S3 bucket name; for ``volume`` this
+        is the on-disk directory the sidecar writes files into.
+        """
+        if result_destination == TransformRequest.OBJECT_STORE_DEST:
+            return request_id
+        if result_destination == TransformRequest.VOLUME_DEST:
+            return os.path.join(
+                TransformerManager.POSIX_VOLUME_MOUNT,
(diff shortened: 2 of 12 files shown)


servicex_app/servicex_app/models.py
```diff
@@ -194,6 +194,7 @@ class TransformRequest(db.Model):
     workers = db.Column(db.Integer, nullable=True)
     result_destination = db.Column(db.String(32), nullable=False)
     result_format = db.Column(db.String(32), nullable=False)
+    output_path = db.Column(db.String(1024), nullable=True)
```
ponyisi, line 197: I guess one might ask is 1024 characters enough... Postgres has no performance hit for unlimited VARCHAR and this is an internal field so a denial of service isn't a concern.

servicex_app/servicex_app/models.py
```diff
@@ -230,6 +231,7 @@ def to_json(self):
             "workers": self.workers,
             "result-destination": self.result_destination,
             "result-format": self.result_format,
+            "output-path": self.output_path,
```
ponyisi, line 234: `to_json` is mostly for the benefit of the client ... is there any reason we would want the client to know this path?

servicex_app/servicex_app/transformer_manager.py
```diff
@@ -79,6 +79,22 @@ def make_api(cls, celery_app):
...
 
+    @staticmethod
+    def compute_output_path(request_id: str, result_destination: str) -> str:
+        """Return the destination the transformer will write results to.
+
+        For ``object-store`` this is the S3 bucket name; for ``volume`` this
+        is the on-disk directory the sidecar writes files into.
+        """
+        if result_destination == TransformRequest.OBJECT_STORE_DEST:
+            return request_id
```
ponyisi, line 90: This we will want to compute at some point, I guess we imagine this method will be able to handle these cases? will we be able to look up the transform parameters from the `request_id` in this static method?

MattShirley: This is to enforce the core output path business logic. It is implemented in `transformation.submit.SubmitTransformationRequest.post` (main point it needs to be calculated at) as well as `create_job_object` (just used to identify the output path for volume mounts). However, it can (and probably should) be removed from `create_job_object`, with the `output_path` moved to `launch_transformer_jobs`/`create_job_object`. That's the general pattern there (to move things down the parameter list starting from an actual `TransformReq` object).

This would allow `compute_output_path` to be moved out of `TransformManager` and into `transformation.submit.SubmitTransformationRequest._compute_output_path`. This would be cleaner and reduce the spread of `compute_output_path` calls (`output_path` should be used directly).

servicex_app/servicex_app/transformer_manager.py
```diff
@@ -79,6 +79,22 @@ def make_api(cls, celery_app):
...
+        if result_destination == TransformRequest.VOLUME_DEST:
+            return os.path.join(
+                TransformerManager.POSIX_VOLUME_MOUNT,
```
MattShirley, line 93: @ponyisi do you think the `POSIX_VOLUME_MOUNT` should be included in the `output_path` here? I suppose someone could have the same file system on a different mount point. Though right now the posix mount value is hardcoded to "/posix_volume".

servicex_app/servicex_app/resources/servicex_resource.py
```diff
@@ -73,43 +71,3 @@ def _get_app_version(cls):
         except PackageNotFoundError:
             return "develop"
 
-    @classmethod
-    def start_transformers(
```
MattShirley: This method is never called and identical to `TransformerManager.start_transformers` (correct entrypoint)

Issue #1539, opened 2026-09-09
Fable repo-wide review


ariostas:
:robot: _AI text below_ :robot:

This is a whole-repo review at `c3c5b5e7` (develop) covering bugs, performance, simplifications, and modernizations. Every finding below was verified by reading the code and, where it mattered, the pinned library sources (flask-restful 0.3.10, SQLAlchemy 2.0.49, kubernetes 25.3.0, minio 7.2.20, celery 5.4, xrootd 5.8.3, servicex-did-finder-lib 3.2.0rc1, servicex-code-gen-lib 1.3, qastle 0.17/0.19). Line numbers refer to the current develop tree.

Test suites were run locally: `transformer_sidecar` passes (41 tests, needs `PYTHONPATH=src`, see S12). `servicex_app` passes 363/369 with the remaining failures caused by resolving newer versions than `poetry.lock` (kubernetes 36 rejects `V1HorizontalPodAutoscaler(spec=None)` in a test fixture) and by the Celery worker fixture needing a broker, so those are environment artifacts, not regressions.

## Summary

The core flows (submit, dataset lookup, file-complete accounting with row locks) are careful, and the Celery/HTTP contracts between app, sidecar, DID finders, and code generators all line up. The problems cluster in four places:

1. **Authorization is inconsistent.** Ownership is checked only in `DeleteTransform`; cancel, status, results, detail, file URLs, the unpaginated list endpoint, and the web detail pages trust any authenticated user, and the session-cookie branch of `auth_required` skips the pending-user check entirely.
2. **Failure paths in the sidecar handshake hang requests silently.** Any exception between `send()` and `confirm()` deadlocks the science container's shell loop, and the `sys.exit(-1)` path acks and drops the Celery message.
3. **Several wired-up code paths cannot execute.** `DELETE /pending` raises `AttributeError`, `"Fatal"` is written into the status enum column, `Api(app, errors=Flask.errorhandler)` disables flask-restful error handling, the default-users bootstrap points at a nonexistent file, three codegen `translate` scripts import nonexistent modules. Tests mock the model methods these depend on, which is why they pass.
4. **Helm/config drift.** `atlasopenmagic` is deployed by default but never added to `VALID_DID_SCHEMES`; documented affinity/tolerations/CVMFS-PVC formats produce invalid manifests; JWT secret keys are hard-coded in the ConfigMap; the secrets examples use key names the deployment does not read.

[...]

PR #1540, opened 2026-09-09
fix(app): repair API code paths that cannot execute

ariostas:

This fixes a group of Flask app code paths that could never succeed: endpoints that dereference `None` or write invalid values into enum columns and always return 500, a query argument that parsed as `True` for every value including `"false"`, and two JSON fields serialized from the wrong source. Tests that mocked away the broken model calls now exercise the real methods against the sqlite test database.

- **B4** · high · [`14bb71cd`](https://github.com/ssl-hep/ServiceX/pull/1540/commits/14bb71cd9061e20ce93897a67bf1121350153dac) · Assign `TransformStatus.fatal` instead of the string `"Fatal"`, return 404 for an unknown request, drop the redundant commit
  - [ ] approve
  - [ ] reject
- **B5** · high · [`793e66ab`](https://github.com/ssl-hep/ServiceX/pull/1540/commits/793e66abe77bc5d6ab9f93d06d6ca77f522b9fba) · `UserModel.delete_all_pending` builds a real query with `db.session.query(cls)`
  - [x] approve
  - [ ] reject
- **B6** · high · [`0c3b1cba`](https://github.com/ssl-hep/ServiceX/pull/1540/commits/0c3b1cba717f29c3b248403567e1bd488b5fa1b6) · `Api(app)` instead of `Api(app, errors=Flask.errorhandler)`, so resource `HTTPException`s return JSON
  - [x] approve
  - [ ] reject
- **B51** · medium · [`4621062b`](https://github.com/ssl-hep/ServiceX/pull/1540/commits/4621062bdbfed0e2ddd3f6354233d4926e956c1d) · 404 for an unknown dataset id, 422 for an unknown dataset on fileset completion, `null` for a missing `last_updated`
  - [ ] approve
  - [ ] reject
- **B52** · medium · [`d049c69c`](https://github.com/ssl-hep/ServiceX/pull/1540/commits/d049c69c829d82cc5fd1d2d938d5346c7a6293f5) · Return 401 when the user record is gone or has no stored refresh token
  - [ ] approve
  - [ ] reject
- **B53** · medium · [`9085871e`](https://github.com/ssl-hep/ServiceX/pull/1540/commits/9085871e468d2804640c1859edd60196d4f21599) · Use `flask_restful.inputs.boolean` for boolean query arguments and a parser default for `show-deleted`
  - [x] approve
  - [ ] reject
- **B56** · medium · [`21c691b6`](https://github.com/ssl-hep/ServiceX/pull/1540/commits/21c691b690d7c8c911fe7d3d51b85ce195bd1f0d) · Emit `null` for a missing `finish-time` and report `x.file_id` as `"file-id"`
  - [x] approve
  - [ ] reject
- **B63** · medium · [`aa38435a`](https://github.com/ssl-hep/ServiceX/pull/1540/commits/aa38435a01b391baaca3f2104b555c428c31fd7f) · Resolve `LOG_LEVEL`, raise `ValueError` on an unknown name, and apply it through `setLevel`
  - [ ] approve
  - [ ] reject

Part of #1539.

### DIFF
--- a/servicex_app/servicex_app/__init__.py
+++ b/servicex_app/servicex_app/__init__.py
@@ -374,7 +374,7 @@ def create_app(
             )
             sys.exit(-1)
 
-        api = Api(app, errors=Flask.errorhandler)
+        api = Api(app)
 
         # ensure the instance folder exists
         try:
--- a/servicex_app/servicex_app/__init__.py
+++ b/servicex_app/servicex_app/__init__.py
@@ -237,7 +237,10 @@ def create_app(
     logstash_port = os.environ.get("LOGSTASH_PORT")
 
     level = os.environ.get("LOG_LEVEL", "DEBUG").upper()
-    app.logger.level = getattr(logging, level, None)
+    numeric_level = getattr(logging, level, None)
+    if not isinstance(numeric_level, int):
+        raise ValueError(f"Unknown LOG_LEVEL: {level}")
+    app.logger.setLevel(numeric_level)
 
     # remove current handlers
     for h in app.logger.handlers:
--- a/servicex_app/servicex_app/models.py
+++ b/servicex_app/servicex_app/models.py
@@ -530,7 +530,11 @@ class Dataset(db.Model):
             "size": self.size,
             "events": self.events,
             "last_used": str(self.last_used.strftime(iso_fmt)),
-            "last_updated": str(self.last_updated.strftime(iso_fmt)),
+            "last_updated": (
+                str(self.last_updated.strftime(iso_fmt))
+                if self.last_updated is not None
+                else None
+            ),
             "lookup_status": self.lookup_status,
             "is_stale": self.stale,
         }
--- a/servicex_app/servicex_app/models.py
+++ b/servicex_app/servicex_app/models.py
@@ -130,7 +130,7 @@ class UserModel(db.Model):
 
     @classmethod
     def delete_all_pending(cls):
-        num_rows_deleted = db.session.query.filter_by(pending=True).delete()
+        num_rows_deleted = db.session.query(cls).filter_by(pending=True).delete()
         db.session.commit()
         return {"message": "{} row(s) deleted".format(num_rows_deleted)}
 
--- a/servicex_app/servicex_app/models.py
+++ b/servicex_app/servicex_app/models.py
@@ -240,10 +240,12 @@ class TransformRequest(db.Model):
             "files-failed": self.files_failed,
             "files-remaining": self.files_remaining,
             "submit-time": str(self.submit_time.strftime(iso_fmt)),
-            "finish-time": str(self.finish_time),
+            "finish-time": (
+                str(self.finish_time.strftime(iso_fmt))
+                if self.finish_time is not None
+                else None
+            ),
         }
-        if self.finish_time is not None:
-            result_obj["finish-time"] = str(self.finish_time.strftime(iso_fmt))
         return result_obj
 
     @classmethod
@@ -468,7 +470,7 @@ class TransformationResult(db.Model):
         return {
             "id": x.id,
             "request-id": x.request_id,
-            "file-id": x.id,
+            "file-id": x.file_id,
             "file-path": x.file_path,
             "s3-object-name": x.s3_object_name,
             "transform_status": x.transform_status,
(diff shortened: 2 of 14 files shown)


ariostas:
All the changes here look reasonable to me, so I'll mark it as ready for review. I can also confirm that the review was easy with the splitting by commits. Each commit message contains a good description, so it's useful to read it.

All the fixes here seem relatively minor to me, but could resolve random crashes.

servicex_app/servicex_app/resources/internal/transform_status.py
```diff
@@ -23,11 +23,15 @@ def post(self, request_id):
             )
 
             submitted_request = TransformRequest.lookup(request_id)
-            submitted_request.status = "Fatal"
+            if not submitted_request:
```
ponyisi, line 26: I suppose it is possible that this code could be hit but it would be a very serious breach of the transformation state machine (it implies updates to a transformation that does not exist). Is swallowing the error and sending a 404 upstream really the best way to handle this?

servicex_app/servicex_app/resources/internal/transform_status.py
```diff
@@ -23,11 +23,15 @@ def post(self, request_id):
...
+                msg = f"Transformation request not found with id: {request_id}"
+                current_app.logger.error(msg, extra={"request_id": request_id})
+                return {"message": msg}, 404
+
+            submitted_request.status = TransformStatus.fatal
             submitted_request.finish_time = datetime.now(tz=timezone.utc)
             submitted_request.failure_description = status["info"]
             submitted_request.save_to_db()
-            db.session.commit()
```
ponyisi, line 30: This should be OK

servicex_app/servicex_app/__init__.py
```diff
@@ -374,7 +374,7 @@ def b64decode(text):  # needed for python selection decoding to web page.
             )
             sys.exit(-1)
 
-        api = Api(app, errors=Flask.errorhandler)
+        api = Api(app)
```
ponyisi, line 377: I feel this will have downstream effects but it is technically the correct thing to do so I will approve

servicex_app/servicex_app/resources/internal/fileset_complete.py
```diff
@@ -51,6 +51,13 @@ def put(self, dataset_id):
         summary = request.get_json()
         dataset = Dataset.find_by_id(int(dataset_id))
 
+        if dataset is None:
```
ponyisi, line 54: Again not sure if catching this inconsistent state and sending back a HTTP error is really what one wants

servicex_app/servicex_app/models.py
```diff
@@ -530,7 +530,11 @@ def to_json(self):
             "size": self.size,
             "events": self.events,
             "last_used": str(self.last_used.strftime(iso_fmt)),
-            "last_updated": str(self.last_updated.strftime(iso_fmt)),
+            "last_updated": (
```
ponyisi, line 533: OK

servicex_app/servicex_app/resources/users/token_refresh.py
```diff
@@ -47,6 +47,9 @@ def post(self):
         claims = get_jwt()
 
         user = UserModel.find_by_email(claims["sub"])
+        if not user or not user.refresh_token:
```
ponyisi, line 50: This should distinguish the two cases. A missing refresh token is something the user can remedy through the web portal so the client should be able to catch this.

servicex_app/servicex_app/models.py
```diff
@@ -240,10 +240,12 @@ def to_json(self):
             "files-failed": self.files_failed,
             "files-remaining": self.files_remaining,
             "submit-time": str(self.submit_time.strftime(iso_fmt)),
-            "finish-time": str(self.finish_time),
+            "finish-time": (
```
ponyisi, line 243: API change but safe downstream

servicex_app/servicex_app/__init__.py
```diff
@@ -237,7 +237,10 @@ def revoke_admin_command(email):
     logstash_port = os.environ.get("LOGSTASH_PORT")
 
     level = os.environ.get("LOG_LEVEL", "DEBUG").upper()
-    app.logger.level = getattr(logging, level, None)
+    numeric_level = getattr(logging, level, None)
+    if not isinstance(numeric_level, int):
+        raise ValueError(f"Unknown LOG_LEVEL: {level}")
+    app.logger.setLevel(numeric_level)
```
ponyisi, line 243: and what happens when this gets set to `None` ?

PR #1546, opened 2026-09-09
fix(app): enforce request ownership and pending-user checks

ariostas:

Tightens authorization in the Flask app: transform requests can now only be read, cancelled or deleted by their submitter or an administrator, session-cookie callers are subject to the same account checks as API token callers, the transform listing endpoint no longer exposes every user's requests, and the Slack interaction handler verifies the request signature before it contacts the caller-supplied `response_url`.

- **B1, B2** · high · [`c031b5b`](https://github.com/ssl-hep/ServiceX/pull/1546/commits/c031b5b3ba9b97da825e1b714243074f4e6364a2) · Restrict transform request endpoints and web pages to the submitter or an administrator, and make the session branch of `auth_required` load the user and apply the same missing and pending checks as the JWT branch; the two findings share one commit because they touch the same functions and each one's tests depend on the other
  - [ ] approve
  - [ ] reject
- **B3** · high · [`7569d3d`](https://github.com/ssl-hep/ServiceX/pull/1546/commits/7569d3d2fe11ada85ad699459e9215ec182b2954) · Scope the transform listing endpoint to the requesting user unless they are an administrator, and stop treating `submitted_by=0` as an absent filter
  - [ ] approve
  - [ ] reject
- **B28** · high · [`3281284`](https://github.com/ssl-hep/ServiceX/pull/1546/commits/3281284d53a7ff0c9e80456f09f1d81ae09686a4) · Verify the Slack HMAC signature before any outbound request to the caller-supplied `response_url`, fix the misparenthesized timestamp expiry check, and answer the unimplemented `reject_user` action with a Slack message instead of an uncaught `NotImplementedError`
  - [ ] approve
  - [ ] reject

- B3 returns 403 when a non-admin asks for another user's `submitted_by` rather than silently rescoping the query to their own requests. Silently rescoping would avoid leaking whether that user has requests, so this is a deliberate choice worth confirming.
- `web/slack_msg_builder.py:missing_slack_app` and `verification_failed` are no longer called anywhere after B28, since those paths must not contact `response_url` before the signature is verified. They were left in place rather than deleted.

Part of #1539.

### DIFF
--- a/servicex_app/servicex_app/resources/servicex_resource.py
+++ b/servicex_app/servicex_app/resources/servicex_resource.py
@@ -25,10 +25,10 @@
 # CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-from typing import Optional
+from typing import Any, Optional, Tuple
 
 from importlib.metadata import version, PackageNotFoundError
-from flask import current_app
+from flask import current_app, session
 from flask_jwt_extended import get_jwt_identity
 from flask_restful import Resource
 from servicex_app.models import UserModel, TransformRequest, TransformStatus
@@ -50,7 +50,7 @@ class ServiceXResource(Resource):
         return "http://" + current_app.config["ADVERTISED_HOSTNAME"] + "/" + endpoint
 
     @staticmethod
-    @jwt_required_if_auth_enabled
+    @jwt_required_if_auth_enabled(optional=True)
     def get_requesting_user() -> Optional[UserModel]:
         """
         :return: User who submitted request for resource.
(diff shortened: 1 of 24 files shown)


servicex_app/servicex_app/resources/transformation/file_urls.py
```diff
@@ -70,12 +69,9 @@ def post(self):
 
         args = parser.parse_args()
         request_id = args["request_id"]
-        # Validate that the user is an admin or submitted the request (not yet implemented)
-        transform = TransformRequest.lookup(request_id)
-        if not transform:
-            msg = f"Transformation request not found with id: {request_id}"
-            current_app.logger.error(msg, extra={"request_id": request_id})
-            return {"message": msg}, 404
+        _, error = self._get_owned_request(request_id)
```
MattShirley, line 72: High priority change

servicex_app/servicex_app/resources/servicex_resource.py
```diff
@@ -59,9 +59,35 @@ def get_requesting_user() -> Optional[UserModel]:
...
         user = None
         if current_app.config.get("ENABLE_AUTH"):
-            user = UserModel.find_by_email(get_jwt_identity())
+            if session.get("is_authenticated"):
+                user = UserModel.find_by_id(session.get("user_id"))
+            else:
+                user = UserModel.find_by_email(get_jwt_identity())
         return user
 
+    def _get_owned_request(self, request_id) -> Tuple[Optional[TransformRequest], Any]:
```
MattShirley, line 68: Remove the error return value and raise exceptions for the 404, 403 errors directly from the method body.

servicex_app/servicex_app/web/transformation_results.py
```diff
@@ -2,13 +2,16 @@
...
 from servicex_app.models import TransformationResult, TransformRequest
+from servicex_app.web.utils import user_owns_request
 
 
 @oauth_required
 def transformation_results(id_: str):
     treq = TransformRequest.lookup(id_)
     if not treq:
         abort(404)
+    if not user_owns_request(treq):
```
MattShirley, line 13: Is it necessary to add `user_owns_request` alongside the new `_get_owned_request`?

PR #1547, opened 2026-09-09
perf(app): index files.dataset_id and remove per-row work in lifecycle tasks

ariostas:

Cuts the per-row and per-request work out of the app's hot paths and bounds its outbound HTTP calls: the two lifecycle endpoints now do their deletions in bulk inside a single transaction, the `files` table gets back the index it needs, authenticated requests decode their JWT once instead of twice, the presigning S3 client is reused, and no outbound request can hang forever.

- **B32** · medium · [`ec37cf6`](https://github.com/ssl-hep/ServiceX/pull/1547/commits/ec37cf6ea90695dbe3a7bc8f5b0b8940bb351ae5) · Give `REQUEST_TIMEOUT` a real read timeout and add one to the VPS liveness request
  - [ ] approve
  - [ ] reject
- **B33** · medium · [`2ea3d89`](https://github.com/ssl-hep/ServiceX/pull/1547/commits/2ea3d892a4e1b9e67acf1722487b27cb6d3ba037) · Remember only images confirmed to exist, so failed registry lookups are retried
  - [ ] approve
  - [ ] reject
- **B55** · medium · [`80ffe44`](https://github.com/ssl-hep/ServiceX/pull/1547/commits/80ffe44ebd6cd80b3d07a6505e78378f6c892409) · Restore the index on `files.dataset_id` and declare the existing `submit_time` index on the model
  - [ ] approve
  - [ ] reject
- **P1** · perf · [`d872baa`](https://github.com/ssl-hep/ServiceX/pull/1547/commits/d872baae0fe7e2b1fc0189ebac1412794aefc935) · Bulk update in dataset obsoletion, one transaction and eager loading in the data lifecycle task
  - [ ] approve
  - [ ] reject
- **P2** · perf · [`15f6c33`](https://github.com/ssl-hep/ServiceX/pull/1547/commits/15f6c33df7f11c4a3a248210120048d935ad9d1d) · Empty buckets with minio's batched `remove_objects` and log through the app logger
  - [ ] approve
  - [ ] reject
- **P6** · perf · [`0e271c1`](https://github.com/ssl-hep/ServiceX/pull/1547/commits/0e271c1282d600944f11ead4ec45c82bec91422d) · Build the presigning S3 client once per set of object store settings and compute the expiry in UTC
  - [ ] approve
  - [ ] reject
- **P7** · perf · [`eae77b5`](https://github.com/ssl-hep/ServiceX/pull/1547/commits/eae77b5a2ebabefe1bc14d8dc420479c4249f782) · Stop verifying the JWT a second time inside `get_jwt_user`
  - [ ] approve
  - [ ] reject

- PR #1549 adds a migration with the same parent revision (`v1_8_4`); whichever of the two merges second needs its `down_revision` re-pointed at the other.
- `servicex_app_test/test_transformer_manager.py` passes only as part of the full suite: `test_init_invalid_config` and `test_persistent_claim_exists` rely on an application context that an earlier test file leaves pushed. This is pre-existing and untouched here, but it makes the file impossible to run on its own.
- `urllib3.Timeout` does not implement `__eq__`, so the VPS liveness test compares `connect_timeout` and `read_timeout` by hand instead of using `assert_called_with`. A small comparison helper would read better if more call sites start passing timeouts.
- The sidecar has the other half of P6: `transformer_sidecar/src/transformer_sidecar/transformer.py` constructs a fresh `ObjectStoreManager`, and therefore a fresh minio client, in two places rather than reusing one.
- `Dataset.find_by_id` in `models.py` still uses the legacy `cls.query.get(id)`, which SQLAlchemy 2.0 deprecates in favour of `db.session.get`; the other two call sites in the same file were already converted.

Part of #1539.

### DIFF
--- a/servicex_app/servicex_app/resources/internal/dataset_lifecycle_ops.py
+++ b/servicex_app/servicex_app/resources/internal/dataset_lifecycle_ops.py
@@ -30,7 +30,7 @@ from datetime import datetime, timedelta, timezone
 from flask import request, current_app
 
 from servicex_app.resources.servicex_resource import ServiceXResource
-from servicex_app.models import Dataset
+from servicex_app.models import Dataset, db
 
 
 class DatasetLifecycleOps(ServiceXResource):
@@ -44,17 +44,20 @@ class DatasetLifecycleOps(ServiceXResource):
         except Exception:
             return {"message": "Invalid age parameter"}, 422
         delta = timedelta(hours=age)
-        datasets = (
-            Dataset.get_all()
-        )  # by default this will only give non-stale datasets
-        todelete = [
-            _.id for _ in datasets if _.last_updated and (now - _.last_updated) > delta
-        ]
+        cutoff = now - delta
+        obsoleted = (
+            db.session.query(Dataset)
+            .filter(
+                Dataset.stale.is_(False),
+                Dataset.last_updated.isnot(None),
+                Dataset.last_updated < cutoff,
+            )
+            .update({Dataset.stale: True}, synchronize_session=False)
+        )
+        db.session.commit()
         current_app.logger.info(
             f"Obsoletion called for datasets older than {delta}. "
-            f"Obsoleting {len(todelete)} datasets."
+            f"Obsoleting {obsoleted} datasets."
         )
-        for dataset_id in todelete:
-            Dataset.delete_dataset(dataset_id)
 
         return {"message": "Success"}, 200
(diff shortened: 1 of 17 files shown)


servicex_app/servicex_app/resources/internal/dataset_lifecycle_ops.py
```diff
@@ -44,17 +44,20 @@ def post(self):
         except Exception:
             return {"message": "Invalid age parameter"}, 422
         delta = timedelta(hours=age)
-        datasets = (
-            Dataset.get_all()
-        )  # by default this will only give non-stale datasets
-        todelete = [
-            _.id for _ in datasets if _.last_updated and (now - _.last_updated) > delta
-        ]
+        cutoff = now - delta
```
MattShirley, line 47: High priority change

servicex_app/servicex_app/resources/transformation/file_urls.py
```diff
@@ -78,7 +83,7 @@ def post(self):
             return {"message": msg}, 404
 
         expirydelta = 365 * 24 * 60 * 60
-        expiry = int(datetime.datetime.now().timestamp() + expirydelta)
+        expiry = int(datetime.now(timezone.utc).timestamp() + expirydelta)
```
MattShirley, line 86: High priority change

PR #1548, opened 2026-09-09
refactor(app): remove the vestigial RabbitMQ adapter and dead code

ariostas:

Removes the vestigial RabbitMQ adaptor from the Flask app and clears out a batch of dead code, duplicated logic and legacy API calls that the repo review turned up. No behavior change is intended beyond two noted cases: the "ServiceX Reports" admin index is now access controlled, and the CLI user create command no longer swallows database errors. A few signatures that were already being edited also picked up PEP 604 unions (M5); there was no repo-wide sweep for that, so it does not get a commit of its own.

- **S1** · simplification · [`2e57dd0`](https://github.com/ssl-hep/ServiceX/pull/1548/commits/2e57dd0b44d78d6c7a47865631aa14572a85fb98) · Delete `rabbit_adaptor.py` and the `rabbitmq_adaptor` plumbing through `create_app`, `add_routes` and `make_api`, plus the `pika` dependency; nothing published to or consumed from those exchanges and queues, and real transform work flows through Celery, which keeps `RABBIT_MQ_URL` as its broker.
  - [x] approve
  - [ ] reject
- **S4** · simplification · [`53202e2`](https://github.com/ssl-hep/ServiceX/pull/1548/commits/53202e2622496ec7a2c8c5ebfa3daf3897b06103) · Drop the uncalled `create_tables()`, the unreachable `return super().format(record)`, the `sys.path.append("/opt/servicex/celery")` for a directory the app pod does not mount, and the lambda wrapper in `_override_config_with_environ`; in `decorators.py` drop the no-op `assert "NoAuthorizationError"` and collapse `except (NoAuthorizationError, Exception)`.
  - [x] approve
  - [ ] reject
- **S5** · simplification · [`53f0d6a`](https://github.com/ssl-hep/ServiceX/pull/1548/commits/53f0d6a5cbca90ce267d71878b982471d0e85f60) · Remove the `try/except NoResultFound` wrappers around `.all()` queries, which cannot fire, make `Dataset.get_by_did_finder` and `Dataset.get_all` return `.all()` in every branch so they match their list annotations, drop a commented-out lookup, and remove the Flask-SQLAlchemy < 3 fallback in `migrations/env.py`.
  - [x] approve
  - [ ] reject
- **S6** · simplification · [`3e7121d`](https://github.com/ssl-hep/ServiceX/pull/1548/commits/3e7121d40ea344fc7bce18708f292b06ada42644) · Drop the `is_complete` rejection in `add_files_to_processing_queue`, which duplicated the live check in `LookupResultProcessor` off a stale JSON snapshot, unify `advertised_endpoint()` with `ServiceXResource._generate_advertised_endpoint`, and move task routing from a `celeryd_after_setup` hook to `task_routes` in `celeryconfig.py`.
  - [ ] approve
  - [ ] reject
- **S7** · simplification · [`ee5ea9d`](https://github.com/ssl-hep/ServiceX/pull/1548/commits/ee5ea9dc3143a983c8a76f6151d4a53a0560d0c8) · Remove the redundant second `isinstance(view, ReportView)` check and give the "ServiceX Reports" admin a secured index view mirroring the main admin; in the results resource use the `to_json_list` classmethod instead of calling `to_json` on an instance and drop an unreachable `request_id` branch.
  - [ ] approve
  - [ ] reject
- **M2** · modernization · [`8d8d38c`](https://github.com/ssl-hep/ServiceX/pull/1548/commits/8d8d38c38bcaadedae60012e5334bdcab114f5ad) · Replace `distutils.util.strtobool`, removed from the standard library in Python 3.12, with a small private helper accepting the same true and false spellings case insensitively.
  - [ ] approve
  - [ ] reject
- **M4** · modernization · [`d46cda5`](https://github.com/ssl-hep/ServiceX/pull/1548/commits/d46cda54a056c376e43bfac2bdcaadcabec5c0bc) · `Dataset.find_by_id` uses `db.session.get(cls, id)` instead of the legacy `cls.query.get(id)`, matching `UserModel.find_by_id` and silencing the `LegacyAPIWarning`.
  - [ ] approve
  - [ ] reject
- **cli** · chore · [`9da66f2`](https://github.com/ssl-hep/ServiceX/pull/1548/commits/9da66f2e43822bb6b886a2f89c50583ffcf3e69f) · Inline `check_user_exists`, a one-line alias for `UserModel.find_by_sub`; `add_user` rolls back and re-raises instead of printing the exception and reporting success; `list_users` builds one query instead of fetching every user and then running a second filtered query.
  - [ ] approve
  - [ ] reject
- **tooling** · chore · [`cd942b4`](https://github.com/ssl-hep/ServiceX/pull/1548/commits/cd942b48a2e06bbde63bc626f8c43ab542c84f66) · Delete `servicex_app/.pre-commit-config.yaml`, which pinned flake8 3.8.4 from a gitlab URL that can no longer be cloned and so broke a repo-wide `pre-commit run -a`. The root config is the one CI uses.
  - [ ] approve
  - [ ] reject

**Skipped**

- S7 `resources/datasets/get_all.py`: the `"show-deleted" in args` check always being true was left alone to avoid colliding with other in-flight work on the resource handlers.
- S2, S3, S13 and the rest of `transformer_manager.py` are out of this PR's scope. `cancel_consumer` in `transformer_manager.py` is a Celery control call, not part of the removed adaptor, so it was left alone.

- `servicex_app/.flake8` excludes three files that no longer exist: `validate_requests.py`, `find_m.py`, `messaging.py`.
- `servicex_app/pyproject.toml` still lists `pre-commit = "^2.20.0"` in the dev group.

Part of #1539.


servicex_app/servicex_app/decorators.py
```diff
@@ -129,7 +128,7 @@ def is_admin_user() -> bool:
         verify_jwt_in_request(locations=["headers"])
         user = get_jwt_user()
         return user is not None and user.admin
-    except (NoAuthorizationError, Exception):
```
ponyisi: This makes the intent more clear but agreed it is redundant

servicex_app/servicex_app/__init__.py
```diff
@@ -323,8 +317,6 @@ def b64decode(text):  # needed for python selection decoding to web page.
             transformer_manager = provided_transformer_manager
 
         if not provided_celery_app:
-            sys.path.append("/opt/servicex/celery")
```
ponyisi: This directory doesn't exist anyway

PR #1551, opened 2026-09-12
Fix copy-paste error in xrootd deployment

ponyisi:
Pull policy for xrootd pointed at wrong variable


BenGalewsky (approved):
Oops
Thank you

PR #1555, opened 2026-09-21
feat(did-finder-rucio): make RSE expression and availability configurable

### LINKED ISSUE #1538: Make Rucio RSE expressions configurable
## Problem
Currently the [Rucio DID Finder](https://github.com/ssl-hep/ServiceX/blob/c3c5b5e7749af04822f6307d83359974d8339b7a/did_finder_rucio/src/rucio_did_finder/rucio_adapter.py#L180) hard codes some ATLAS Rucio RSEs in the query 
```
rse_expression=r"istape=False&(*\type=SPECIAL|cloud=CERN)"
```

These don't work for CMS. Make them a configurable setting via Values.yaml

BenGalewsky:
The Rucio DID finder hardcoded the RSE expression and ignore_availability flag passed to list_replicas, which only suited the ATLAS production deployment. Both are now plumbed through LookupRequest from the RUCIO_RSE_EXPRESSION and RUCIO_IGNORE_AVAILABILITY environment variables, exposed in the helm chart as didFinder.rucio.rseExpression and didFinder.rucio.ignoreAvailability. Defaults preserve the previous behavior.

Fixes #1538 

# Testing
I've deployed this to prod since it was automatically deployed as part of the helm chart workflow. I updated the rucio-did-finder image to match.

Everything works as normal. I changed the RSE Expression to
```
        rseExpression: 'istape=True&(*\type=SPECIAL|cloud=FOOBAR)'
````

And when I delete the dataset and force a lookup I see this in the log files
```
servicex_did_finder_lib.exceptions.LookupFailureException: Dataset user.rjanusia:CalRatio_Emulation_Pileup_AOD_v2 is missing replicas for 1 of its files. extra: {'dataset_id': 193}
```

Reverting to the ATLAS RSE and it works again.

### CODE REVIEW

File: docs/deployment/reference.md [hunk 1]
```diff
@@ -61,6 +61,8 @@ parameters for the [rabbitMQ](https://github.com/bitnami/charts/tree/master/bitn
...
 | `didFinder.rucio.rucio_host`                 | URL for Rucio service to use                                                                                                                                                                                                       | `https://voatlasrucio-server-prod.cern.ch:443` |
 | `didFinder.rucio.auth_host`                  | URL to obtain Rucio authentication                                                                                                                                                                                                 | `https://atlas-rucio-auth.cern.ch:443`         |
+| `didFinder.rucio.rseExpression`              | Rucio RSE expression used to filter which replicas are returned for a dataset                                                                                                                                                      | `istape=False&(*\type=SPECIAL\|cloud=CERN)`     |
```
ponyisi: do we want to leave a comment that the default is appropriate for ATLAS ... ?

did_finder_rucio/src/rucio_did_finder/lookup_request.py
```diff
@@ -40,6 +40,8 @@ def __init__(
         dataset_id: str = "bogus-id",
         replica_sorter: Optional[ReplicaSorter] = None,
         location: Optional[Mapping[str, float]] = None,
+        rse_expression: str = DEFAULT_RSE_EXPRESSION,
```
ponyisi: I feel that embedding these defaults is probably going to cause confusion down the road - it's at best redundant with what's happening upstream in the `values.yaml`. I personally would just have the DID finder die with an obvious error if somehow `RUCIO_RSE_EXPRESSION` isn't available in the environment.

ponyisi (approved):
Since I guess this is an urgent PR, I'll approve, but see my comments - I'm uncomfortable with having experiment-dependent default fallbacks embedded so deeply in the code when we should "for sure" be passing them as arguments.

docs/deployment/reference.md
BenGalewsky: Yes - Fixed

did_finder_rucio/src/rucio_did_finder/lookup_request.py
BenGalewsky: Indeed - I was working under the model of "For god's sake, don't break the Chicago AF", but that should be handled in values.yaml and this is better

BenGalewsky:
Your wish is Claude's commad

